PR #39, opened 2014-12-04
Changing modelName to type

ksykulev:
modelName does not appear to be in scope in the pushPayload function.
It appears that this is a typo for type which is the first argument to
the pushPayload function.


danielspaniel:
@ksykulev, I just did a test and this style worked for me .. the one like:

```js
var cars = FactoryGuy.makeList('car', 2, 'with_manufacturer', 'with_parts');
testHelper.handleFindQuery('car', ['year', 'manufacturer'], cars)
```
 I made sure to make the associations ( parts, and manufacturer ) embedded as well.
 So, can you give it a try and let me know if that works for you too?  If not, can you send me the tests your trying ( and the setup ) .. so I can compare and see what I am doing differently.

danielspaniel:
@ksykulev , Is {embedded:'always'} even supported anymore in association definition. I know its in use in the serializer definitions( as of 1.0.0.beta.10), so it makes me wonder if the other is even relevant? Can you confirm that.

ksykulev:
@danielspaniel - Hey sorry for the delayed response. Here is some tests I came up with: 
https://gist.github.com/ksykulev/998583603bb8df423f81

Let me know which ones are completely out of bounds and which ones make sense. I can definitely use the pattern of usage you recommend for handleFindQuery.

RE: embedded - good question. I need to look more deeply into this. 

Unfortunately i've been really busy at work and haven't had a chance to regroup about the fixture adapter changes or this new embedded question. I'll try to allocate some time soon to both of these things.

danielspaniel:
No problem @ksykulev. You know, now that I think about it and see your tests, I am thinking of removing the 'passing in json' altogether, and just have the array you pass in be records already created with make or makeList ( third parameter to handleFindQuery ).

 Do you care?  Is there any great reason to pass in json fixtures. I only supported that because you were using it that way, but the reason to not support it anymore is because I am not sure how reliable it is, whereas I can almost guarantee that passing in records will work every time.

ksykulev:
@danielspaniel - I think it's totally reasonable to not support JSON with handleFindQuery for now/until we find a good reason for it. If the user needs that fine grained control they should just use mockjax/sinon on their own. no? The could just as easily use FactoryGuy.build(list) to help them create the JSON as well..

danielspaniel:
Ok @ksykulev, I will drop support for the json array ( of fixtures ), so you will have to make the records you want to be returned by the handleFindQuery before you call it.  We can always bring it back if there is need. 
 I will push out the new version tomorrow. ( 0.9.3 )

ksykulev:
@danielspaniel - With the last round of changes things have been working really smoothly. Thanks for your help.

danielspaniel:
Your welcome @ksykulev  :)

PR #47, opened 2015-01-10
Fixture Adapter Tests Pass

ksykulev:
Reworking what FactoryGuy.make does when using the fixture adapter.
It now not only mutates the FIXTURE array on the corresponding model but also updates the store with the changes.

For instance:
```
var user = FactoryGuy.make('user');
store.all('user').get('length') //1. Contains user.
```

With this change the fixture adapter + factory guy more closely mirrors how the non-fixture adapters + factory guy functions. 

Let me know if this change is something you want to add to factory guy.


ksykulev:
I just realized that if you agree with this change we will have to slightly modify the documentation. Most notably in this section:
https://github.com/danielspaniel/ember-data-factory-guy#using-dsfixture-adapter

danielspaniel:
Right @ksykulev .. can you update the docs?
Thanks for the fixture adapter updates btw.

danielspaniel:
@ksykulev, I think you are using browser to run tests?  
When I ran 'grunt test' ( which travis does when running CI ) .. some of the tests did not run, so I changed them slightly to make them work with 'grunt test'.  
You might notice that when you pull in my changes. It's annoying, that phantomjs runs tests differently than browswer, but I have gotten used to it.
 So, what I am saying is just make sure to run 'grunt test' when you are all set.

ksykulev:
@danielspaniel - Hmm weird. I usually run grunt test before I commit. And they seemed to pass locally.
```
$ grunt test
Running "concat:dist" (concat) task
File dist/ember-data-factory-guy.js created.

Running "qunit:all" (qunit) task
Testing tests/index.html ..............................................................................................................................OK
>> 329 assertions passed (3289ms)

Done, without errors.
```

Sorry about that.

danielspaniel:
Oh yeah .. It's not a huge deal, just tripped me up a few times, so alerting you.

danielspaniel:
@ksykulev .. I just updated to ember-data.1.0.0-beta.15 and some of the fixture adapter tests are breaking .. I pushed out tag version 0.9.9 anyway .. and commented out the fixture-adapter tests .. 
 do you feel like taking a look and seeing if you can fix those?

ksykulev:
@danielspaniel - hmm interesting. I will take a look and see what I can do.

danielspaniel:
thanks @ksykulev .. I was pleasantly surprised that all the other tests passes since the last beta version was breaking lots of them.

ksykulev:
@danielspaniel - Just wanted to give you an update... I haven't forgotten about this, just still haven't had a chance to take a look at it deeply. 
But at first (and quick) glance led me to the test_helper file. In particular the way the container and the store is being set up..

danielspaniel:
thanks for the update @ksykulev  .. and if you see better way to set up the tests .. container/store etc .. feel free to change that setup.

PR #50, opened 2015-01-14
Add handleFindOne method

Marthyn:
to handle get requests with a specific id.

like

```
/api/v1/profiles/1
```

will return a record with id 1.


danielspaniel:
@Marthyn .. can you give me use case for the handleFindOne .. it's interesting because I think I used to have that method, but I took it out a long time ago because I thought people could do:
```
FactoryGuy.make('user');
store.find('user', user.id).then(function(user) { // blah });
```
for this case.

Marthyn:
True, but for example a reload on that object will still trigger an api request. I found in our app that even though we created the object with factoryguy there were still requests occurring, which we used to catch with pretender but we want to switch.

danielspaniel:
Oh boy .. totally forgot about reload .. got it.

danielspaniel:
Are you using bower? If so you can get this code by updating since its now in master branch. If not (using gem) let me know and I will make new release version.

danielspaniel:
@Marthyn .. and thanks for updating the readme documentation ( almost no one does that )

Marthyn:
We're using the gem :)

danielspaniel:
v0.9.4 gem  is now ready

Marthyn:
Thanks! Another use case i found btw is this:

`self.transitionToRoute('chapter', self.get('model.id'));`

you transition to the id of an object, instead of to the object itself. So it forces a reload.

danielspaniel:
Interesting .. I have never done that in a transition .. only passed model .. good to know.

PR #55, opened 2015-01-19
Allows custom mapping of objects.

pedrokiefer:
Allows one to subclass / reopen FactoryGuyTestMixin and configures a custom object mapping.

For example, Django Tastypie returns:
```
objects: [ { }, { }, ... ]
meta: { ... }
```

FactoryGuyTestMixin only returns:
```
clients: [ ... ]
```


danielspaniel:
@pedrokiefer, Did not realize that was issue, so thanks for bringing that up and making pull request for it.  This code is checked in to master, so will work if you using bower to just reload v0.9.6. Let me know otherwise and I will make new patch version.

pedrokiefer:
@danielspaniel great! thanks for the fast response :)

PR #59, opened 2015-01-23
Missing var in mockjax handler

mmelvin0:
This was causing requests mocked with handleCreate() (and probably others) to fail in 0.9.7, whereas they were working in 0.8.4.


danielspaniel:
@mmelvin0 , can you give me an example of a failed test you have ( that this new code fixes )  so I can recreate that and add it to my test suite.

danielspaniel:
I think you have a good point here and the line should read: 

```
for (var attribute in expectedRequest) {
```

instead of 
```
for (attribute in expectedRequest) {
```

ie. the var should be in the for .. in statement .. 

which I will change .. so thank you for pointing that out, and I am still curious which tests you have fail without this var ( though jshint complains rightfully )

mmelvin0:
I think this may be because I'm using Ember CLI which stamps ['use strict'](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Converting_mistakes_into_errors) on everything.

For me, the error was simply "ReferenceError: attribute is not defined" whenever mockjax was indirectly invoked by my acceptance test after having called handleCreate().

I'm pretty sure it would fail for any mocked request.

I'm also very curious what the unit test for this would look like. I took a look at your tests briefly but couldn't figure out how I would go about testing what mockjax was doing, directly or indirectly. I'm very new to JavaScript unit testing/qUint so I'm interested in seeing how testing interactions with a wrapper library works.

danielspaniel:
@mmelvin0 , maybe I could try using 'use strict' in my code, or running jshint ( which would have caught that one ) .. this would mean I don't necessarily need to test this problem, since those two solutions would be showing me the problem right away. 
 Also, let me know if you need a new version to pull in the new code ( with the var correctly placed ) ..

Issue #60, opened 2015-01-30
Issue when using handleUpdate


pedrokiefer:
I'm writing some tests using handleUpdate, but it keeps failing. I think it's because factory-guy returns an empty set on PUT requests. 

The error message is the following:
"An adapter cannot assign a new id to a record that already has an id. <my-app@model:mymodel::ember882:1> had id: 1 and you tried to update it with null. This likely happened because your server returned data in response to a find or update that had a different id than the one you sent."

In this project I'm using ember-data-tastypie-adapter, ember 1.9.1 and ember-data beta.14. Maybe returning the id would solve, or serializing the hole object, not sure though. Maybe this specific for this adapter, or related to beta.14.

danielspaniel:
True, @pedrokiefer  about the return of empty for PUT's.  I put a notice on the README saying I was having problems with the beta.14.1 release and was going to wait for beta.15 to come out to see if that was better ( whacky issues that seemed ember data related ) .. so my question is, can you revert to ember-data beta.12 for now ( that is what I am doing ) .. and it should  be fine.

pedrokiefer:
@danielspaniel sure I can. It's just hard to keep up with ember folks, there is always a new release and lots of new features. I haven't read the changelog for what is going on on beta.15, so I can't comment if this issue will go away. 

My idea to solve this issues was to serialize the object and return it. The problem is, we need to serialize it just before returning the data and not on handleUpdate registration, or else the update will return the old object data.

danielspaniel:
Exactly @pedrokiefer ... not so simple .. and in the ember beta.12 release returning empty is the standard  when nothing changes except the updates you made, so that strategy was pretty good. 
 For the next beta release I will have to adjust and as you say .. I may have to do something tricky to get that to work ..  not sure yet though, but in the next few weeks when that is released I will try and fix it.

pedrokiefer:
I was reading mockjax docs, and it accepts a function for response and calls it before returning data. So we can have something like:

```
function create_response_callback(modelName, record, mapping) {
    return function (settings) {
        var json = record.toJSON({includeId: true});
        var responseJson = mapping(modelName, json);
        this.responseText = responseJson;
    }
}
```

Where modelName is the model name, record is the record being updated and mapping is ```FactoryGuyTestMixin.mapFind``` function. 

```stubEndpointForHttpRequest``` would need to accept a the result of this function and set mockjax options to use this function.

[...]

PR #63, opened 2015-02-19
[WIP] Handle update

pedrokiefer:
This is a basic implementation of what we discussed on #60. I'm open to suggestions, but I think this is a good starting point.


danielspaniel:
wow .. @pedrokiefer .. I am glad you finally got around to this .. give me a day to look it over .. but so far, looks like good start.

danielspaniel:
Ok @pedrokiefer ... first question .. did you write a test for the problem this new code solves. 
  I am alittle unclear how/what this is doing differently ( except for the chainable methods part ( which is nice ) ) than the old code.  I guess I am not sure what this does different than before is what I am saying.

pedrokiefer:
It solves the issue I was having with ember-data 14.1, tests in the project I'm working were **failing** because the mocked ajax was **not returning data**. After updating to my branch, all tests passed.

Right now it should not do anything else as the old code, expect returning a json object on the mocked ajax call. But this new code allows us to do things like custom failure codes with different response text, mock a timeout, and maybe some other scenarios. I'll add tests and use cases for this shortly.

pedrokiefer:
The error I was getting is the following:

```
Error: Assertion Failed: An adapter cannot assign a new id to a record that already has an id. <my-project@model:something::ember895:1> had id: 1 and you tried to update it with null. This likely happened because your server returned data in response to a find or update that had a different id than the one you sent.
    at new Error (native)
    at Error.EmberError (http://localhost:4200/assets/vendor.js:22910:23)
    at Object.Ember.assert (http://localhost:4200/assets/vendor.js:14407:15)
    at Ember.Object.extend.updateId (http://localhost:4200/assets/vendor.js:78982:15)
    at Ember.Object.extend.didSaveRecord (http://localhost:4200/assets/vendor.js:78933:16)
    at http://localhost:4200/assets/vendor.js:79736:17
    at Object.Backburner.run (http://localhost:4200/assets/vendor.js:10667:27)
    at ember$data$lib$system$store$$_adapterRun (http://localhost:4200/assets/vendor.js:79568:32)
    at http://localhost:4200/assets/vendor.js:79730:9
    at tryCatch (http://localhost:4200/assets/vendor.js:60908:16)
```

After updating I don't get any errors and my tests passes :)

danielspaniel:
Ok @pedrokiefer, thanks for adding more tests (  if all the possibilities are covered I'd be happy ) .. and perhaps you can fix up the documentation a show these changes to handleUpdate ?? and I will merge this.

pedrokiefer:
@danielspaniel just added the new docs for handleUpdate, please review it :)

danielspaniel:
Me like :) .. thanks for being thorough with docs and tests

PR #66, opened 2015-02-24
Use  typeKey for relationship instead of attribute name for build fixture relationship

cristinawithout:
When building a fixture with relationships, FactoryGuy.build() is called with the relationship attribute name rather than that attributes model type.

It currently works only when relationship attributes have the same name as the model type they're related to.

So this works:
User = DS.Model.extend({
    company:     DS.belongsTo('company',)
});

But this doesn't:
User = DS.Model.extend({
    default_company:     DS.belongsTo('company',)
});

Build tries to find a factory named 'default_company' rather than the 'company' factory.

### DIFF
--- a/src/factory_guy.js
+++ b/src/factory_guy.js
@@ -62,14 +62,15 @@ var FactoryGuy = {
    @param {String} attribute  attribute you want to check
    @returns {Boolean} true if the attribute is a relationship, false if not
    */
-  isAttributeRelationship: function(typeName, attribute) {
+  getAttributeRelationship: function(typeName, attribute) {
     if (!this.store) {
       Ember.debug("FactoryGuy does not have the application's store. Use FactoryGuy.setStore(store) before making any fixtures")
       // The legacy value was true.
       return true;
     }
     var model = this.store.modelFor(typeName);
-    return !!model.typeForRelationship(attribute);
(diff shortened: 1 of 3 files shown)


danielspaniel:
Thanks Christina for fixing that up. Can't believe I did not see that before. If you want me to make new version that has this change let me know and I will bump to v0.9.11

cristinawithout:
Thanks. Up to you if you want to bump the version. We can point to the SHA until there's a new version.

danielspaniel:
@cristinawithout .. version 0.9.11 was released a few days ago .. fyi 
 also, I noticed that you have made the ember-cli-factory-guy  package an ember-addon. 
  2 questions. 
How ( when using as add on ) .. do you update the version of ember-data-factory-guy?
 I it better to keep your project or make mine into an addon ( basically merging what your project has into mine ) 
 I was just about to embark on making mine into an addon when I noticed you already did that. Anyway .. alittle confused since there are so many ways to use factory-guy.  I just include it in my ember-cli bower.json and others use your addon ( adding to package.json ) ( which I think is great by the way ) ..   so I thought it might be nice to have just one package ( ember-data-factory-guy )  and then use it this way or that .. if you know what I mean.

PR #78, opened 2015-04-29
Get this add-on working when prototype extensions are disabled in Ember

chrisdpeters:
The app I'm working on requires that I [disable prototype extensions](http://guides.emberjs.com/v1.10.0/configuring-ember/disabling-prototype-extensions/). Factory Guy wasn't coded to accommodate that.

This commit at least gets the tests passing after I changed `tests/dummy/config/environment.js` so that `EXTEND_PROTOTYPES` is `false`. I'm not sure that it fixes all of the issues that may be introduced, but at least it's letting me test my app now.

The [new default](https://github.com/ember-cli/ember-cli/issues/3443) for developing add-ons is to have prototype extensions turned off, so this brings Factory Guy up to code with that standard.

### DIFF
--- a/addon/factory-guy-test-helper.js
+++ b/addon/factory-guy-test-helper.js
@@ -91,7 +91,7 @@ var FactoryGuyTestHelper = Ember.Object.create({
   */
   mapFindAll: function(modelName, json) {
     var responseJson = {};
-    responseJson[modelName.pluralize()] = json;
+    responseJson[Ember.String.pluralize(modelName)] = json;
     return responseJson;
   },
   /**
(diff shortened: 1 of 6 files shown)


danielspaniel:
Interesting @chrisdpeters, I think it makes sense to support as wide a variety of setups as possible, so thanks for the PR.

chrisdpeters:
You're welcome. I see that you tagged 1.0.6 on GitHub but have not yet released the update on NPM. I'm getting antsy here. :)

danielspaniel:
Thanks for reminder @chrisdpeterss, I keep forgetting I have to publish the npm version. 1.0.6 is now published.

PR #83, opened 2015-05-08
ignore autogenerated .jscs files

GlennGeelen:
When adding Broccoli-JSCS to our Ember project, Ember-data-factory-guy breaks our app because it doesn't filter the auto generated JSCS files. This only occurred with the following setting for broccoli-jscs in the broccoli.js.
```
jscsOptions: {
  enabled: true,
  disableTestGenerator: false
}
```

With this PR the Ember app also ignores the auto generated files made by Broccoli-JSCS.


danielspaniel:
Your reminding me that I did not like that setup in the first place .. need to find a better way to import the factory files and only the factory files.  Can you think of a better way to do it ( off the top of your head ) before I merge this?

GlennGeelen:
No not right now but I'll think about it, maybe I'll come up with something.

danielspaniel:
Ok .. I am going to ponder it again too .. want me to merge this and make new patch version in the meantime?

GlennGeelen:
Yes please.

danielspaniel:
new version 1.0.8 is ready

GlennGeelen:
Thanks

Issue #84, opened 2015-05-13
add .andFail() option to handleFind()


alexanderjeurissen:
I'm currently writing a integration spec for a ember-rails project where I want to test if our `notfound` a.k.a 404 route is triggered when a resource is requested that does not exist.

this is what I have so far:
```javascript
test('renders not-found page for nonexisting ids', function () {
  expect(1);

  visit('/products/1');

  testHelper.handleFind('product', {id: "1"});

  andThen(function () {
    expectRouteToBe('not_found'); //custom helper that checks if currentPath() equals X
  });
});
```

The problem is that I do want to mock the `/products/1` endpoint BUT i want it to fail and return a 404. As far as I'm aware off this functionality isn't implemented as of this moment (but I could be wrong).

When I look at the API documentation I see that there is a `.andFail` option for `handleCreate()` and `handleUpdate()` it would be nice to have this option for the `handleFind()` helper as well.

I'm willing to create a PR for this (including specs), as long as you are interested in this functionality / use-case.

danielspaniel:
Sure thing @alexanderjeurissen .. that is a good idea.  If you make the PR .. I'll use it.

danielspaniel:
version 1.0.10 has andFail method added to handleFind

PR #85, opened 2015-05-19
[WIP] Add andFail() method to handleFind()

alexanderjeurissen:
This pullrequest implements issue #84 it adds the same `andFail()` method chaining functionality that is already implemented for `PUT` and `DELETE` requests to the `handleFind()` test-helper.

TODO:

- [ ] fix specs
- [ ] update comments to reflect new method

### DIFF
--- a/addon/mock-get-request.js
+++ b/addon/mock-get-request.js
@@ -1,0 +1,51 @@
+import $ from 'jquery';
+
+var MockGetRequest = function (url, modelName, record, mapFind) {
+  var status = 200;
+  var succeed = true;
+  var response = null;
+
+  this.andSucceed = function (options) {
+    succeed = true;
+    status = options && options.status || 200;
+
+    return this;
+  };
+
+  this.andFail = function (options) {
+    console.log('INFO: andFail is called');
+    options = options || {};
+    succeed = false;
+    status = options.status || 500;
+    response = options.response || {};
+
+    return this;
+  };
+
+  this.handler = function () {
+
+    if (succeed === false) {
+      this.status = status;
+      this.responseText = response;
+    } else {
+      this.status = status;
+      this.responseText = mapFind(
+        modelName,
+        record.toJSON({
+          includeId: true
+        })
+      );
+    }
+  };
+
+  var requestConfig = {
+    url: url,
+    dataType: 'json',
+    type: 'GET',
+    response: this.handler
+  };
+
+  $.mockjax(requestConfig);
+};
+
+export default MockGetRequest;
--- a/tests/unit/factory-guy-test-helper-test.js
+++ b/tests/unit/factory-guy-test-helper-test.js
@@ -3,6 +3,7 @@ import FactoryGuy, { make, makeList } from 'ember-data-factory-guy';
 import TestHelper from 'ember-data-factory-guy/factory-guy-test-helper';
 import MissingSequenceError from 'ember-data-factory-guy/missing-sequence-error';
 
+import $ from 'jquery';
 import User from 'dummy/models/user';
 import BigHat from 'dummy/models/big-hat';
 import SmallHat from 'dummy/models/small-hat';
@@ -274,6 +275,24 @@ test("#handleFind with traits and arguments", function (assert) {
   store.find('profile', 1).then(function (profile) {
     ok(profile.get('description') === description);
     done();
+  });
+});
+
+test("#handleFind failure with andFail method", function (assert) {
+  Ember.run(function () {
+    var done = assert.async();
+    TestHelper.handleFind('profile', 1).andFail();
+
+    store.find('profile', 1).then(
+      function (response) {
+        ok(false);
+        done();
+      },
+      function (error) {
+        ok(true);
+        done();
+      }
+    );
   });
 });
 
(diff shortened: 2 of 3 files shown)


alexanderjeurissen:
@danielspaniel somehow the specs are failing, I'm not sure why but the error callback of the promise isn't resolved.

```console
not ok 105 PhantomJS 1.9 - FactoryGuyTestHelper with DS.RESTAdapter : #handleFind failure with andFail method
    ---
        actual: >
            false
        expected: >
            true
        message: >
            failed, expected argument to be truthy, was: false
        Log: >
            INFO: requestConfig set
            INFO: andFail is called
            RESPONSE: {"created_at":null,"description":"Text goes here","camelCaseDescription":null,"snake_case_description":null,"company":null,"group":null}
    ...

As you can see the response is as what we'd expect in case the `andFail()` was not present. so somehow the succeed flag isn't set properly, or the handler is called before the `andFail()` method has set succeed to false.

I'm having the feeling that I miss something very obvious, could you weigh in on this ?

File: addon/mock-get-request.js [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+
+  this.andSucceed = function (options) {
+    succeed = true;
+    status = options && options.status || 200;
+
+    return this;
+  };
+
+  this.andFail = function (options) {
+    console.log('INFO: andFail is called');
```
alexanderjeurissen: ugh forgot this one.. will remove it.

tests/unit/factory-guy-test-helper-test.js
```diff
@@ -3,6 +3,7 @@ import FactoryGuy, { make, makeList } from 'ember-data-factory-guy';
 import TestHelper from 'ember-data-factory-guy/factory-guy-test-helper';
 import MissingSequenceError from 'ember-data-factory-guy/missing-sequence-error';
 
+import $ from 'jquery';
```
alexanderjeurissen: same here, added this for debugging purposes. will remove in next commit.

alexanderjeurissen:
@danielspaniel that won't do, because somehow the promise resolves the success callback, without the `done()` statement in the success callback the test will hang, resulting in the following error:

```shell
not ok 105 PhantomJS 1.9 - FactoryGuyTestHelper with DS.RESTAdapter : #handleFind failure with andFail method
    ---
        actual: >
            null
        message: >
            Test timed out
        Log: >
    ...
```

danielspaniel:
right .. well, the problem then seems to be that the request should fail, and it's not, which means maybe your request mock is not catching the request? can not exactly tell without running this test myself and checking it out.  If you can't figure it out today, I will work on it tomorrow.

alexanderjeurissen:
yeah seems that for whatever reason the request is not catched by `$.mockAjax` I'll look into it some more, when I'm stuck I'll let you know

danielspaniel:
Hi @alexanderjeurissen .. I am working on this now, and I see the problem. 
 You could help me a ton by telling be your use cases for handleFind  ( success and failure ) , since I may change this functionality around based on how it is really used. 
  I can see it for testing reload .. of model you have.  But for the rest of the cases I am a bit unclear.

alexanderjeurissen:
Hi @danielspaniel , I've been busy lately and haven't found time to fix this PR. I'm curious as to what the problem was ?

I'll gladly provide you with some use-cases / specs where we are currently using the `handleFind` method.

**usecase 1 (handleFind success):**

```javascript
test('create new path and redirect', function () {
  expect(1);

  visit('/paths/new');
  fillIn('.path-header > .inner > input', 'Javascript Ninjas');
  fillIn('.path-header > .inner > textarea', 'Become a ninja at JavaScript');

  click('button.button-save');

  testHelper.handleCreate('path');
  testHelper.handleFind('path', {
    id: 1
  });

  andThen(function () {
    expectRouteToBe('path.index');
  });
});

```

this spec uses the `handleFind` to make sure the store contains a `path` because after saving a new `path` the app redirects to the index route to show the newly created path.

**usecase 2 (handleFind failure):**

```javascript
test('renders not-found page for nonexisting ids', function () {
  expect(1);

  visit('/products/1');

  testHelper.handleFind('product', {id: "1"});

  andThen(function () {
    expectRouteToBe('not_found'); //custom helper that checks if currentPath() equals X
  });
});
```

here we want to test if our router correctly handles redirecting to a `not_found` aka `404` route in case a resource is requested that does not exist.

danielspaniel:
Thanks @alexanderjeurissen .. these examples were very helpful.  What I see is that I have made some mistakes in how I assumed handleFind would be used. I was creating a model and placing it in the store in the handleFind method ( which is why the ajax call is never made.)
  So, for your first example that would actually be bad, since you don't want another model, you want the one you just created. 
  But in that case actually .. you don't need to say handleFind after a handleCreate, since the store will ( after the handleCreate/ or createRecord/ save  events will now have that model you just created. 
  No need to handleFind, since the store already knows it has it and will NOT do an ajax call to get it. 
  Therefore you don't need that handleFind  ( correct me if I am wrong .. or you find that you need it ) 
 
 The second case is what I figured would be the traditional one, and knowing that .. I will now fix the and fail to work in these traditional cases. 
   
  I was trying to make handleFind too fancy before, and I will strip it down to be more basic.

danielspaniel:
@alexanderjeurissen, version 1.0.10 has the andFail method working

PR #89, opened 2015-06-09
Mocking a failed reload typo fix

nummi:
Removed extra right curly brace


danielspaniel:
thanks for the fix @nummi,  good catch.

nummi:
Thanks for the library!

PR #96, opened 2015-06-19
Add “then” callback for handleCreate

dweebit:
Whatcha think about adding a callback for the async helpers that get called after the mockjax response is sent? I can add tests and extend it to the other helpers if you like it.

### DIFF
--- a/addon/mock-create-request.js
+++ b/addon/mock-create-request.js
@@ -10,6 +10,7 @@ var MockCreateRequest = function(url, store, modelName, options) {
   var responseJson = {};
   var expectedRequest = {};
   var modelType = store.modelFor(modelName);
+  var thenCallback = function() {};
 
   this.calculate = function() {
     if (matchArgs) {
@@ -43,6 +44,10 @@ var MockCreateRequest = function(url, store, modelName, options) {
     returnArgs = returns;
     this.calculate();
     return this;
+  };
+
+  this.then = function(callback) {
+    thenCallback = callback;
   };
 
   this.andFail = function(options) {
@@ -83,6 +88,7 @@ var MockCreateRequest = function(url, store, modelName, options) {
       this.status = status;
     }
     this.responseText = responseJson;
+    thenCallback(responseJson);
   };
 
   var requestConfig = {


danielspaniel:
Hmm .. @dweebit,  can you give me a use case so I can picture it better .. and for sure, would need some tests

dweebit:
Sure thing! My main personal motivation in having something like this is to be able to interrogate the mocked out response payload. My team used to have a lot of code that would do something like this:

```javascript
var myThingId = 42;
TestHelper.handleCreate("thing", {returns: {id: myThingId}});
triggerCreationOfThing();
TestHelper.handleUpdate("", myThingId);
triggerAnUpdateOnMyThing();
```

Where the only reason we were passing in an explicit ID into the handleCreate was so we'd have a way to identify the newly created object after the fact and test it further. 

However, since the addition of this https://github.com/danielspaniel/ember-data-factory-guy/blob/master/addon/mock-create-request.js#L80, our explicitly set IDs are blown away right before the response is created. Soooo, if we had a 'then' callback that gets passed the response payload, then we can pull it apart to get the ID and write further tests based off of it.

```javascript
TestHelper.handleCreate("thing")
.then(function(response) {
  var newThingId = response['thing']['id'];
  TestHelper.handleUpdate('thing', newThingId);
  triggerAnUpdateOnMyThing();
});
triggerCreationOfThing();
```

Thinking about it further, it might be more appropriately named beforeResponse considering its current implementation.

danielspaniel:
I saw this in mockjax docs: https://github.com/jakerella/jquery-mockjax#performing-actions-after-request-completion

 and also .. maybe I should just fix that issue of the id being blown away .. and should set that id that you want in the returns?  Would that seem reasonable?

dweebit:
Fixing the ID being blown away would work. Yet another possible solution would be a public method to peek at the next model ID for a factory; something like a non-destructive version of [nextId()](https://github.com/danielspaniel/ember-data-factory-guy/blob/bfb898c9a8e2b69e0f840c2d04d1d9b47aa65db2/addon/model-definition.js#L33).

The main reason I went the route of a callback with my pull request was that I figured it'd have a broader application than my specific use-case, that it could handle multiple open handleCreate requests better, and it would allow factory guy to be totally in control of making unique IDs.

So, we've found three possible solutions for my problem. Which of them would you want implemented? I'm happy to turn this pull request in whatever you want it to be :)

danielspaniel:
That use case is pretty whacky .. to manipulate the response in a callback ..  you know .. if you can pull it off I will buy you a beer ( though I might not merge it ) , because I don't think it's going to be easy to do. And I still can't see the usefulness of it. ( I keep looking at your example and just don't follow why you would do that , update a response while it in progress of being mocked ?  )

 To find the next thing created is also not too bad .. just do something like: 
 ```
 var newlyCreatedThing = store.all('thing').get('content.lastObject') 
 after you do the action that created the thing

 I think fixing the issue of the id being blown away would be my choice .. If you can submit pull request I will merge that one in first.

dweebit:
Sorry I wasn't clear . . . the motivation for passing the response object into the callback function is not to modify it, but to read values from it. But it's ok, I can drop it :)

I'll submit a separate pull request for the ID thing.

dweebit:
Since https://github.com/danielspaniel/ember-data-factory-guy/pull/98 was merged I wont take it the wrong way if you trash this pull request. However if you want any of the ideas herein to be implemented, I'd be happy to help.

danielspaniel:
I am going to close this for now, but if you get a hankering to talk about this again, or have other ideas let me know.

PR #105, opened 2015-07-02
add note to avoid using ember-data-1.0.0-beta.19.2

raycohen:
surfacing useful info from https://github.com/danielspaniel/ember-data-factory-guy/issues/94#issuecomment-112493918


danielspaniel:
Good catch @raycohen , and sorry you had to dig through all the documentation to realize that you had to avoid 19.2 .. what a pain. But now that that is all settled .. happy factorying. :)

PR #106, opened 2015-07-13
Fix DEPRECATION: Ember.keys is deprecated in favor of Object.keys

jcope2013:
introduced in ember 1.13.3


danielspaniel:
Yes, good catch @jcope2013, and thank for the PR. 
I should have been using Object.keys in the first place.

PR #114, opened 2015-07-27
Work without prototype extensions

indirect:
Ember addons are expected to work without prototype extensions, because an application may or may not have prototype extensions enabled, and addons need to work with any application. (Plus, my application has them disabled, and Factory Guy broke.)

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -470,6 +470,7 @@ var FactoryGuy = function () {
     Add the model to included array unless it's already there.
    */
   var addToIncluded = function (included, data) {
+    included = Ember.A(included);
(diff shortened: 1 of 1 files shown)


danielspaniel:
@indirect, I will pull this in for the next version hopefully tomorrow

indirect:
:+1: :raised_hands:

danielspaniel:
one thing to note is that this was included in version 1.13.3 .. which might mean you will have to upgrade to ember data 1.13.5.  I am hoping it is not much of an inconvenience.

indirect:
We're already on ember-data 1.13.7, so that's all good for us. :)

PR #119, opened 2015-08-11
Ember 2.0 fixes

mcm-ham:
* Fix error due to removal of DS.ActiveModelAdapter
* Fix error due to container.lookup being undefined in initializer


danielspaniel:
thanks @mcm-ham for the PR :)

Issue #116, opened 2015-08-06
JSON API attribute name conversion when using a custom application serializer


begedin:
In our application, we're using ember data with JSON API on the client side, Rails API with Active Model Serializers on the server side.

ActiveModelSerializers, by default, give us attributes in underscored form, so we had to override our application serializer like so:

```JavaScript

import Ember from 'ember';
import DS from 'ember-data';

var underscore = Ember.String.underscore;

export default DS.JSONAPISerializer.extend({
  keyForAttribute: function(attr) {
    return underscore(attr); // needed to transform AMS's underscores
  },

  keyForRelationship: function(rawKey) {
    return underscore(rawKey); // needed to transform AMS's underscores
  }
});

```

However, when we define a factory, we have to use camelized properties:

```JavaScript
FactoryGuy.define('user', {
 default: {
   admin: false,
   canTrial: false,
   email: 'johndoe@gmail.com',
   fullName: 'John Doe',
   id: 1
 }
});
```

From debugging, I'm seeing that what factory guy does is the following:

1. Get the factory
2. Run it through the default JSON API serializer to get a **dasherized** form
3. Pass it on to the store as an adapter payload
4. The store then uses the actual application serializer (our custom one), to convert back to the client camelized form.
5. Since the adapter expects an underscored form as input, the multi-word properties simply get ignored.

We could modify our adapter to account for this behavior and simply support both underscored and dasherized form of attribute keys, but I'm thinking FactoryGuy should use the application serializer anyway. Otherwise, it would prevent us from customizing the way our attribute (or relationship) keys are named.

Am I missing something, or is this really what's going on here?

danielspaniel:
@begedin, you are almost 100% correct and missing very little, and I was waiting for someone to bring up this issue, because I knew it was sitting there waiting for someone to pour hot syrup all over it. 
  Unless you mean in item 2 that the 'default JSON API serializer' I am using it factory guy's own serializer, in which case you are 100% correct. 
  What should be happening is that I should use the application's serializer as you are suggesting. 
  Is there any chance that you could fix this and submit a PR? 
 I am on vacation starting tomorrow with not much internet for 10 days, otherwise I would give it a go. 

 In the JSONAPIAttributeTransformer class 

```javascript
var JSONAPIAttributeTransformer = function (/*store*/) {
  var defaultTransformFn = ''.dasherize;
```
I kind of set myself up to override the transform function .. ( sort of ) ... so that might be one angle, 
[...]

PR #122, opened 2015-08-19
Allow usage of applications serializer to transform attributes

begedin:
This should resolve issue #116, but it's just an initial set of commits.

However, if I'm going in the completely wrong direction, anyone is free to point it out.

At this point, It seems the fix works with our project, but it's not covered with tests yet. I haven't quite figured out the tests yet and I don't want to upset the established conventions.

The way I see it, I should

* Have two models which use two serializers, both of which use something other than the default dasherize approach for serializing attribute and relationship keys. Underscore is probably easiest.
* Check that making/building fixtures using those models and their factories gives me models with proper keys and relationships present.

Anything I may have missed?

An unfortunate consequence of the approach I used is that I need to pass in the model name all the way from the top to the bottom (to transformObjectKeys). If anyone has a better idea there, please let me know. In the meantime, I'll keep looking.

### DIFF
--- a/addon/jsonapi-attribute-transformer.js
+++ b/addon/jsonapi-attribute-transformer.js
@@ -62,7 +62,7 @@ var JSONAPIAttributeTransformer = function (store) {
     var transformFunction = serializer.keyForAttribute || Ember.String.dasherize;
     for (var key in object) {
       var value = object[key];
-      var newKey = transformFunction.call(key);
+      var newKey = transformFunction(key);
       delete object[key];
       object[newKey] = value;
     }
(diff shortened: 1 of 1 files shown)


addon/jsonapi-attribute-transformer.js
```diff
@@ -47,21 +46,23 @@ var JSONAPIAttributeTransformer = function (/*store*/) {
...
     if (fixture.attributes) {
-      transormObjectKeys(fixture.attributes);
+      transformObjectKeys(modelName, fixture.attributes);
     }
   };
 
-  var transormObjectKeys = function(object) {
+  var transformObjectKeys = function(modelName, object) {
+    var serializer = store.serializerFor(modelName);
+    var transformFunction = serializer.keyForAttribute || Ember.String.dasherize;
```
begedin: Simply put, we now use the serializer for a specific model name and it's `keyForAttribute` function. I still fallback to `dasherize` by default, but that should never happen, since there's always at least the default serializer the app uses.

addon/jsonapi-attribute-transformer.js
```diff
@@ -71,20 +72,32 @@ var JSONAPIAttributeTransformer = function (/*store*/) {
...
+  var transformRelationshipObjectKeys = function(modelName, object) {
+    var serializer = store.serializerFor(modelName);
+    var transformFunction = serializer.keyForRelationship || Ember.String.dasherize;
+    for (var key in object) {
+      var value = object[key];
+      var newKey = transformFunction(key);
+      delete object[key];
+      object[newKey] = value;
+    }
+  };
```
begedin: This is pretty much a duplicate of `transformObjectKeys`, except it uses `serializer.keyForRelationship`.

danielspaniel:
@begedin, this is right on, and just what I would have done. I will add the tests and get this version published. Thanks for getting it done.

begedin:
Great news! Thank you for your feedback. I'm looking forward to the next release.

PR #129, opened 2015-09-03
update README for handleFindAll

### LINKED ISSUE #128: handleFindAll not playing well with handleUpdate
I'm trying to write integration tests for an index page.  One test involves updating one of the items on the index page.  Basically I haven't been able to get this to work.:

```js
TestHelper.handleFindAll('connection', 1); 
//trying to handle a update of the connection that was returned from the findAll
TestHelper.handleUpdate('connection', 1); 
```

handleFindAll doesn't seem to put the records in the store so when handleUpdate calls 
```js 
model = store.peekRecord(type, id);
```
 it returns null and errors out later on.  

I have instead tried using makeList to create the records, which works with handleUpdate but then my tests fail when the route fails to load all the 'connections' because the query has not been mocked out.

I have also tried to pass the result of makeList  to handleFindAll but that also doesn't seem to work.

Am I missing something simple here, or is there another strategy I should be using?  Or is this a bug?

It would be nice if handleFindAll loaded the records it creates into the store so that we could interact with them afterwards.  

I just updated to version 2.0.1 of FactoryGuy and using ember and ember-data 2.0.0.  I was experiencing the same issue in version 1.13 of FactoryGuy as well.

eliotpiering:
and added test case demonstrating using handleFindAll and handleDelete.  This would close #128.

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,7 +646,7 @@ will look like this:
 
 ##### handleFindAll
   - For dealing with finding all records of a particular type
-  - [Sample acceptance test using handleFindAll (users-view-test.js):](https://github.com/danielspaniel/ember-data-factory-guy/blob/master/tests/acceptance/users-view-test.js)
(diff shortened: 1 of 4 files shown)


danielspaniel:
thanks a ton @eliotpiering .. I really like good documentation .. so that was super great.  Nice test too.

Issue #143, opened 2015-10-28
How should I handle a full scenario involving handleCreate and store.query?


jagthedrummer:
I'm trying to test a full UI flow that starts with no categories, adds one, and then checks to make sure that it is displayed. I'm running into trouble when I try to use `store.query` in my route instead of `store.findAll`.

If my route uses `store.findAll`, then this acceptance test will work perfectly.

```javascript
test('create a new category', function(assert) {
  TestHelper.handleFindAll('category', []);
  visit('/admin/categories');
  
  andThen(function() {
    assert.equal(find('.empty-block').length, 1);
    assert.equal(find('table tbody tr').length, 0);
  });

  click('a:contains(New Category)');

  andThen(function() {
    assert.equal(currentPath(), 'admin.categories.new');

    fillIn('div:contains(Name) input', 'Test Category');
    fillIn('div:contains(Description) input', 'This is only a test!');

    TestHelper.handleCreate('category');
    click('button:submit');
  });

  andThen(function() {
    assert.equal(currentPath(), 'admin.categories.index');
    assert.equal(find('.empty-block').length, 0);
    assert.equal(find('table tbody tr').length, 1);
  });
});
```

But if I change my route to use `store.query` and change the first line of the test to be `TestHelper.handleQuery('category', {}, []);` then the final two `assert.equal` tests fail, and when I `return pauseTest();` before them I can see that as far as Ember is concerned I still don't have any categories.

Am I missing something?

danielspaniel:
What version of ed and factory guy are you using?  
Can you upgrade to 2.1.0? by any chance?

jagthedrummer:
I'm on ED 1.13.14 and EDFG 1.13.10. Does 2.1.0 play nicely with ED 1.13.14?

danielspaniel:
it does .. give it a try ( might not fix issue right off ) .. but then I can work from there to find solution

jagthedrummer:
Cool! I'll give it a shot and report back. Thanks for the quick replies!

jagthedrummer:
OK, I'm seeing the same thing on 2.1.0.

danielspaniel:
good .. 
did you change 
```  TestHelper.handleFindAll('category', []); ```
to 
```  TestHelper.handleFindAll('category', 0); ```

 in the test

danielspaniel:
without stepping through this I can't quite tell though what is the issue. 
  Can you see why the handle create is not creating .. 
[...]

PR #144, opened 2015-10-28
Add some tests to illustrate an issue with handleFindQuery (Issue #143)

jagthedrummer:
This is just a quick and dirty import of the category acceptance test
from my project. I pulled in everything that I needed to make the other
parts of it pass, just to sanity check that things were working.


danielspaniel:
closing this because I think the issue was resolved ( and tested )

PR #145, opened 2015-11-01
Attribute typing

wismer:
@danielspaniel I encountered a potential bug in factory guy - If a definition or a build (with custom attribute) contains an attribute with a default value that is `false`, that attribute does not get carried over into the store or get converted properly. 

```javascript
 // addon/rest-fixture-converter.js
  var extractAttributes = function (modelName, fixture) {
    var attributes = {};
    var transformFunction = getTransformFunction(modelName, 'Attribute');
    store.modelFor(modelName).eachAttribute(function (attribute) {
      var attributeKey = transformFunction(attribute);
      //console.log('attribute', attribute, attributeKey, fixture.hasOwnProperty(attribute), fixture.hasOwnProperty(attributeKey))
      if (fixture[attribute]) {
        attributes[attributeKey] = fixture[attribute];
      } else if (fixture[attributeKey]) {
        attributes[attributeKey] = fixture[attributeKey];
      }
    });
    return attributes;
  };
```

Whichever key is used, `attribute` or `attributeKey`, a value that is `false` will not copy with the conditionals set above. This only seems to affect the REST adapter. In order to make the tests pass, I had to slightly change some of the tests to include the `aBooleanField`. Please let me know if the changes would work or if there's something that I am missing.

### DIFF
--- a/addon/rest-fixture-converter.js
+++ b/addon/rest-fixture-converter.js
@@ -78,10 +78,9 @@ var RestFixtureConverter = function (store) {
     var transformFunction = getTransformFunction(modelName, 'Attribute');
     store.modelFor(modelName).eachAttribute(function (attribute) {
       var attributeKey = transformFunction(attribute);
-      //console.log('attribute', attribute, attributeKey, fixture.hasOwnProperty(attribute), fixture.hasOwnProperty(attributeKey))
-      if (fixture[attribute]) {
+      if (fixture.hasOwnProperty(attribute)) {
         attributes[attributeKey] = fixture[attribute];
-      } else if (fixture[attributeKey]) {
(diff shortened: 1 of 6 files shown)


danielspaniel:
Thanks for the fixer upper .. did you mean to test boolean being false or true .. not sure why you did the flip the bool field intead of adding another field so you would have:
```
  aBooleanField: false,
  bBooleanField: true
``` 

I am going to add that to the tests .. if you know what I mean

danielspaniel:
or just testing both false and true .. is what I meant

wismer:
oh! I should have thought of just testing out both false and true. Yea, that would be best.

PR #148, opened 2015-11-04
Fix jquery-mockjax bower package addition to Project

anilmaurya:
I am using ember-cli@1.13.8, `jquery-mockjax` didn't added automatically to my bower.json after installing ember-data-factory-guy, I had to remove initialize function from `blueprints/ember-data-factory-guy/index.js` to make it work.

I also removed `blueprintsPath` from `index.js` because this is only required when `blueprints` is not present in root directory. http://www.ember-cli.com/extending/#blueprint-conventions


danielspaniel:
ok .. thanks for PR .. I wonder how this ever worked before .. puzzler .. but I will test this out on ember 2 and see if it still works

PR #149, opened 2015-11-11
Don't assume application adapter.

abuiles:
This is not ready and is meant to show an issue which happens when using custom adapter.

If you are using a custom adapter/serializer for some models, the
generated output for build won't be correct since it always assume the
application adapter.

In my example, my application adapter if using "active-model" but I'm
starting to migrate some model to JSONAPI which use a different adapter.

A better approach could be to have a "cache" of fixtureBuilders which is resolve the correct adapter in flight.

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -8,6 +8,7 @@ var FactoryGuy = function () {
   var store = null;
   var fixtureBuilderFactory = null;
   var fixtureBuilder = null;
+  var fixtureBuilders = {};
   var self = this;
   /**
    ```javascript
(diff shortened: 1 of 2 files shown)


addon/factory-guy.js
```diff
@@ -79,6 +80,15 @@ var FactoryGuy = function () {
...
     return fixtureBuilder;
   };
+  this.getFixtureBuilderFor = function (modelName) {
+    if (!fixtureBuilders[modelName]) {
+      var factory = new FixtureBuilderFactory(store, modelName);
+
+      fixtureBuilders[modelName]  = factory.getFixtureBuilder();
+    }
+
+    return fixtureBuilders[modelName];
```
abuiles: A better map could be to have an instance per type instead of model, so if the adapter is json it returns something like `fixtureBuilder['JSONAdapter']`

danielspaniel:
Good points .. and thanks for getting this started ..

danielspaniel:
any progress on this .. it is kind of a nice thing to have .. but requires a bit of testing. let me know if you still working on it?  don't need it? moved away and forgot the whole thing?

juwara0:
I have run into the case where my application level adapter is different from my model level adapter and found this PR that looks like it adds this functionality.

danielspaniel:
yes .. this PR stalled .. but is worth finishing if you want to take a crack at it. ?

abuiles:
@danielspaniel thanks for you work. I stopped using this addon, so I won't be able to continue work on this PR. I'm closing it so others can take on it. 

cc @juwara0

danielspaniel:
@juwara0 .. you want to work on this then?

juwara0:
@danielspaniel - let me reach out to you on slack tomorrow so we can talk about it more.

PR #153, opened 2015-12-04
Removes legacy getStore method

xcambar:
Continuing the work initiated in 7cf90b1c3c7a130298675386fe13edd50477ed78, this PR removes the last occurrences of `FactoryGuy.getStore()`.

### DIFF
--- a/addon/factory-guy-test-helper.js
+++ b/addon/factory-guy-test-helper.js
@@ -25,9 +25,7 @@ var FactoryGuyTestHelper = Ember.Object.create({
     var controller = this.controllerFor(controller_name);
     controller.set(property, value);
   },
-  getStore: function () {
-    return FactoryGuy.get('store');
-  },
+
   /**
    Using mockjax to stub an http request.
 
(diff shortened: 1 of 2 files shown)


danielspaniel:
Thanks again .. I also changed all the internal code ( tests ) to no longer use FactoryGuy.getStore() so we can probably remove that method in a few months as well.

xcambar:
I'm not sure I understand.

The `getStore` method is already gone from all the code.

danielspaniel:
Sorry .. yeah .. I realized that what you did was add to what I started after your PR merged.
 After the merge I changed all internal code to use get('store') and you added to that to change the TestHelper code ( which I forgot ) .. anyway.. 
  whatever happened .. was good .. so I am happy

danielspaniel:
and I thought we could get rid of FactoryGuy.getStore if other people using it ( but they probably are not )

xcambar:
> whatever happened .. was good .. so I am happy

That's the most important :smile: 

Thanks for clarifying.

PR #154, opened 2015-12-05
No more useJSONAPI

xcambar:
In order to be compatible with custom formatters, removing `FactoryGuy.useJSONAPI()` was necessary, mostly because there was no way to tell FactoryGuy how to fetch IDs (see https://github.com/danielspaniel/ember-data-factory-guy/compare/master...xcambar:no_more_useJSONAPI?expand=1#diff-163b229985a285775137075bd760cff9L123).

`useJSONAPI` was used for 3 reasons:

1. extract the IDs from a payload (see above)
1. determine which HTTP verb to use for updates
1. wraps the payload (for RESTSerializer)

The responsibility for these has been transfered to the Builder, which are now the only objects aware (and needing to be aware) of the expected format.

Hope you'll like it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Ember Data Factory Guy  
+# Ember Data Factory Guy
 
 [![Build Status](https://secure.travis-ci.org/danielspaniel/ember-data-factory-guy.png?branch=master)](http://travis-ci.org/danielspaniel/ember-data-factory-guy) [![Ember Observer Score](http://emberobserver.com/badges/ember-data-factory-guy.svg)](http://emberobserver.com/addons/ember-data-factory-guy) [![npm version](https://badge.fury.io/js/ember-data-factory-guy.svg)](http://badge.fury.io/js/ember-data-factory-guy)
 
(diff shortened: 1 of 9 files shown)


danielspaniel:
Pretty darn slick .. I must say. Nice refactor .. and a much needed one there. I was getting tired of the useJSONAPI method ... big time.  And you sliced it away and still kept the 'mapFind' ability .. :) 

 Question? 
  Why do this:
```
  updateHTTPMethod: Ember.computed('fixtureBuilder', {
    get() {
      return this.getWithDefault('fixtureBuilder.updateHTTPMethod', 'PUT');
    }
 }),
```

 instead of this:

```
  updateHTTPMethod: Ember.computed('fixtureBuilder', {
      return this.getWithDefault('fixtureBuilder.updateHTTPMethod', 'PUT');
   }),
```

me confused .. seems like the get() part is redundant ?

xcambar:
Thank you for your comments!

About the syntax, I've just used the new Computed Properties syntax introduced in Ember 1.12.
See http://emberjs.com/blog/2015/05/13/ember-1-12-released.html#toc_new-computed-syntax and http://guides.emberjs.com/v2.2.0/object-model/computed-properties/ for details.

danielspaniel:
Right .. I am aware of this new syntax .. the question comes from the statement in the docs:

In Ember 1.12, you can nicely separate the getter and setter into two different functions.

 and in some of your computed properties ( like the one I mentioned ) .. there is no setter override, so it is cleaner ( seems to me ) to not use the full on get() {} format ... since that is the default.  Know what I mean?

danielspaniel:
Anyway ..  thanks for all the help .. that was a great commit.

xcambar:
You're right, I've read again the docs, and as you've pointed out, the rule seems to be:

* no setter: function
* setter: get/set object.

Good to know for future code.

PR #161, opened 2015-12-09
handleFindAll() and buildList() can make diverse models

afn:
The `handleFindAll` function is a bit lacking in that all of the models it generates are identical (except for sequences).  This PR allows it to be called with a list of traits/options instead of a number.

```javascript
// findAll('user') will return Alice, Bob, and Charlotte (who is an admin)
TestHelper.handleFindAll('user', {name: 'Alice'}, {name: 'Bob'}, ['admin', {name: 'Charlotte'}]);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -745,6 +745,19 @@ will look like this:
   - For dealing with finding all records of a particular type
   - Sample acceptance tests using handleFindAll: [(users-view-test.js)](https://github.com/danielspaniel/ember-data-factory-guy/blob/master/tests/acceptance/users-view-test.js) [(users-delete-test.js)](https://github.com/danielspaniel/ember-data-factory-guy/blob/master/tests/acceptance/users-delete-test.js)
 
+Usage:
+
+```javascript
+  // The mock API can return several different models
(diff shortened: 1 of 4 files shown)


afn:
That's right: if the second argument is not a number, then it's assumed that the second and subsequent arguments are the specification for each model.  Each of these arguments can be an array `['trait1', 'trait2', {key: 'value'}]`, an object `{key: 'value'}`, or a single trait `'trait1'`.

danielspaniel:
This clearly is a fun feature. So thanks for the great idea. Can you add a few more tests ( you have one I think ) .. need one or two for handleFindAll  and then change the README to show the new feature ..

afn:
@danielspaniel How does that look?

danielspaniel:
Pretty good mate .. thanks! :)

afn:
Thanks for merging this so quickly! Are you planning a new release soon?

danielspaniel:
Yes, I will do this today ( I was wondering when you would ask .. so thanks for the nudge ). Was waiting to do other upgrades, but I am taking too long.

PR #162, opened 2015-12-10
Deprecates options succeed

xcambar:
A small one, I saw your comment [here](https://github.com/danielspaniel/ember-data-factory-guy/compare/master...xcambar:deprecates_options_succeed?expand=1#diff-9872310e9a623729710540206c140e36R59) about supporting old non-chaining methods and thought that could be good to use `Ember.deprecate` to help developers migrate, or at least show them that it is deprecated.

FYI, I indicated version `3.0.0` as the limit of availability because it will be a breaking change, and I simply followed SemVer here.


danielspaniel:
I have used the Ember.deprecate before .. and I like that new thing about when the version will kill the feature .. nice one.

PR #167, opened 2015-12-23
Return fixture relationship IDs if not an object (fix #166)

### LINKED ISSUE #166: FactoryGuy.make() doesn't handle ID relationships, Ember Data does
If I do this (with parent.children being a hasMany relationship to child):

```
FactoryGuy.make('child', {id: 1});
FactoryGuy.make('parent', children: [1]);
```

`JSONAPIFixtureConverter.extractRelationships()` returns undefined for these relationships.

However, Ember Data itself will look up these IDs and replace them with the child records, if FactoryGuy wasn't replacing the IDs with undefined. Based on the Ember Data behavior - and this worked for our project with Factory Guy 0.9.1 - it seems simple enough to fix.

I'll prepare a pull request, let me know if you have further insight into this.

mmelvin0:
Handle FactoryGuy.make() with relationships specified as IDs.


danielspaniel:
Sorry I did not include this in last release @mmelvin0. For some reason I just missed it. Ugh.

PR #169, opened 2015-12-31
Supporting nested query params comparisons in handleQuery

gleseur:
Hi,

Migrating from 1.13.10 to 2.1.3 I noticed that handleQuery no longer works for nested query parameters

   TestHelper.handleQuery('foo', { a: { b: 'c' } })
   ...
   this.store.query('foo', { a: { b: 'c' } })

from https://github.com/danielspaniel/ember-data-factory-guy/commit/fcce319a96262cbfbd63ad9b28d73e60d977436f I'd say it's not on purpose.

This adds support for this case that is not so uncommon.

### DIFF
--- a/addon/mock-query-request.js
+++ b/addon/mock-query-request.js
@@ -13,6 +13,11 @@ function isEquivalent(a, b) {
 
   for (var i = 0; i < aProps.length; i++) {
     var propName = aProps[i];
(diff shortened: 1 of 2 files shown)


danielspaniel:
Thanks for the PR ( with test ), good catch.

PR #173, opened 2016-01-22
build(): apply transform of custom type when serializing attributes

kemenaran:
Ember-Data models can have attributes defined with a custom type, which [applies a transform](https://guides.emberjs.com/v2.0.0/models/defining-models/#toc_transforms) when the attribute is serialized or deserialized.

For now, Ember-Data-Factory-Guy's `make()` already support these kind of custom attributes. But `build()` just ignores the transform, and puts the raw attribute value in the generated payload.

This MR makes `RESTFixtureConverter` and `JSONAPIFixtureConverter` aware of transformed attributes.

```c
// model/user.js
export DS.Model.extend({
  info: attr('object'),
});

// transform/object.js
export DS.Transform.extend({
  serialize: function(deserialized) {
    return deserialized ? JSON.stringify(deserialized) : "{}"
  }
});

// Tests
build('user', {info: {first: 1}} );
var expectedJson = {
  user: {
    id: 1,
    info: '{"first":1}' // info is now correctly serialized as a string (instead of being a hash)
  }
};
```

### DIFF
--- a/addon/jsonapi-attribute-transformer.js
+++ b/addon/jsonapi-attribute-transformer.js
@@ -8,6 +8,7 @@ import $ from 'jquery';
  * @constructor
  */
 var JSONAPIAttributeTransformer = function (store) {
+  var defaultValueTransformFn = function(x) { return x; };
 
   /**
    * Transform attributes in fixture.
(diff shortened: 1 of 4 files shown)


danielspaniel:
Thanks for helping and and fixing this up, and for adding the tests :)

kemenaran:
@danielspaniel awesome, thanks for looking at this!

danielspaniel:
version 2.1.4  was just published and has this fix included in the release.

PR #185, opened 2016-04-04
Fix bug with model fragments support

mattmcginnis:
See: https://github.com/danielspaniel/ember-data-factory-guy/pull/184#issuecomment-205084674

### DIFF
--- a/addon/converter/fixture-converter.js
+++ b/addon/converter/fixture-converter.js
@@ -39,7 +39,8 @@ export default class {
     if ('function' === typeof serializer.transformFor) {
       // Use serializer.transformFor for compatibility with ember-data-model-fragments,
       // for context see https://github.com/danielspaniel/ember-data-factory-guy/issues/182
-      return serializer.transformFor(type).serialize;
+      let fragmentType = serializer.transformFor(type);
+      return fragmentType.deserialize.bind(fragmentType);
     }
     let container = Ember.getOwner ? Ember.getOwner(this.store) : this.store.container;
     return container.lookup('transform:' + type).serialize;


addon/converter/fixture-converter.js
```diff
@@ -39,7 +39,8 @@ export default class {
     if ('function' === typeof serializer.transformFor) {
       // Use serializer.transformFor for compatibility with ember-data-model-fragments,
       // for context see https://github.com/danielspaniel/ember-data-factory-guy/issues/182
-      return serializer.transformFor(type).serialize;
+      let fragmentType = serializer.transformFor(type);
+      return fragmentType.deserialize.bind(fragmentType);
```
whatthewhat: Should this be `return fragmentType.serialize.bind(fragmentType);`?

danielspaniel: Actually I am wondering if this should be a do nothing ( return (x)=>x ) kind of thing.

mattmcginnis: >Should this be return fragmentType.serialize.bind(fragmentType);?

Tried that and got "Cannot read property 'modelName' of undefined" on the last line of this snippet.

```
    /**
      Changes serializer fallbacks for fragments to use `serializer:-fragment`
      if registered, then uses the default serializer.
       @method serializerFor
      @private
      @param {String} modelName the record to serialize
      @return {DS.Serializer}
    */
    serializerFor: function serializerFor(modelOrClass) {
      var modelName;

      if (typeof modelOrClass === 'string') {
        modelName = modelOrClass;
      } else {
        modelName = modelOrClass.modelName;
      }
```

mattmcginnis: Interestingly, if I just return this.noTransformFn, all my tests pass in the sample repo.

danielspaniel: Right, that is what I was saying about the (x)=> x .. function = noTransformFn. 
 I think that is the correct answer .. but I want to get more tests in there and if you can help with that .. then rock on. 
  Need to know more about MF transformations and when they are used.

danielspaniel:
This issue should now be fixed with version 2.5 ( which has full support for ember-data-model-fragments )

PR #187, opened 2016-04-12
Handle active model formatted request data properly

ccleung:
convert the root key from `my-model-name` to `my_model_name` so that we access the data properly.

See issue https://github.com/danielspaniel/ember-data-factory-guy/issues/165 for more detail

### DIFF
--- a/addon/mocks/mock-create-request.js
+++ b/addon/mocks/mock-create-request.js
@@ -103,8 +103,12 @@ let MockCreateRequest = function (url, modelName, options) {
     if (!expectedData.data) {
       expectedData = store.normalize(modelName, expectedData);
     }
-    if (!requestData.data && requestData[modelName]) {
-      requestData = store.normalize(modelName, requestData[modelName]);
+    if (!requestData.data) {
+      const serializer = store.serializerFor(modelName);
+      const transformedModelKey = serializer.payloadKeyFromModelName(modelName);
+      if (requestData[transformedModelKey]) {
+        requestData = store.normalize(modelName, requestData[transformedModelKey]);
+      }
     }
     let expectedAttributes = expectedData.data.attributes;
     let requestedAttributes = requestData.data.attributes;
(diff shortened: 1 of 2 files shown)


danielspaniel:
This was good idea .. 
Overall I am a bit sad about mockCreate .. it needs some lovin' .. even more than you have given it .. so if you have any other good ideas .. send em my way!
One more thing, this does pass tests .. but do you you think you could add one to show that this works so it does not get broken in the future?

ccleung:
@danielspaniel added tests in updated commit

danielspaniel:
with added tests, the build is failing for some reason .. can you check it out?

ccleung:
yeah it looks like DS.RESTAdapter test is failing because the root key is `camelCase` instead of `under_score`

danielspaniel:
I thought that might happen .. if it's not easy to fix I can help out later today.

ccleung:
seems like there is no difference between ActiveModelAdapter and RESTAdapter now? https://github.com/danielspaniel/ember-data-factory-guy/blob/master/addon/builder/fixture-builder-factory.js , im trying to figure out how to differentiate the two cases...

danielspaniel:
They both do the same thing, but they are calling out to get a tranformer for an attribute ( lets say ) .. and that will depend on the adapter type, but the two fixture builders don't have to know what that is. So, for you, you need the payload key ( or something like that ) .. so maybe try and ask the fixtureBuilder ( add a method ? )  because it has the adapter all setup, and then  maybe adapter.payloadKey(modelName)  might do it.  ??

ccleung:
yeah ill try that thanks

ccleung:
Doesn't look like the adapter class has any methods for figuring out the root key, seems like something i can put into fixtureBuilder classes themselves, but that means that we may need to introduce an ActiveModelFixtureBuilder instance, that will have a 'payloadKey' that works differently from RESTFixtureBuilder, would that be okay? ActiveModelFixtureBuilder will extend RESTFixtureBuilder and the only different would be the implementation of the new `payloadKey` method

danielspaniel:
Sorry .. I was wrong about adapter, it's serializer

http://emberjs.com/api/data/classes/DS.RESTSerializer.html#method_payloadKeyFromModelName

ccleung:
@danielspaniel updated, thanks for the feedback and help

danielspaniel:
Thanks for hanging in there! :)

PR #190, opened 2016-04-20
Add changelog

Robdel12:
I used a gem to generate this (and use this in my addons too): https://github.com/skywinder/github-changelog-generator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,782 @@
+# Change Log
+
+## [Unreleased](https://github.com/danielspaniel/ember-data-factory-guy/tree/HEAD)
+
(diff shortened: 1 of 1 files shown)


danielspaniel:
That was pretty interesting .. never thought of that for auto generating changelog .. but do I then have to update it? I usually update the changes in the release notes.

Robdel12:
You can continue to do both. If you run `github_changelog_generator` in your addon directory it'll use the git info and pull together another changelog :)

danielspaniel:
ok .. great .. good to know

Robdel12:
I'm sure there's a nice way to automate it but I haven't dug into it yet

PR #191, opened 2016-04-21
refactor the equivalence function so that it works with arrays as well

bryanaka:
Currently, array parameters do not work for `mockQuery` because this tests strict equivalence versus the contents of the array.

This PR addresses the issue by breaking the `isEquivalent` function into three pieces and then makes use of recursion to deal with nested equality.

The three pieces are: 

1. Object Equivalence - slightly modified take on your algorithm 
2. Array Equivalence - contents and order of `arrayA` are identical to contents and order of `arrayB`
3. Normal Strict Equivalence `===` 

I would like some feedback on how you would like me to test the array case, since `user-search-test` is using `fillIn` and we need an array and an object as a param.

### Left to Do
- [ ] Match the JS Style to your repo's style
- [x] Tests that deal with nested objects and arrays

### DIFF
--- a/addon/utils/helper-functions.js
+++ b/addon/utils/helper-functions.js
@@ -1,28 +1,36 @@
 import Ember from 'ember';
 
-// compare to object for loose equality
-let isEquivalent = function(a, b) {
-
-  let aProps = Object.keys(a);
-  let bProps = Object.keys(b);
-
-  if (aProps.length !== bProps.length) {
-    return false;
+function isEquivalent(a, b) {
+  var type = Ember.typeOf(a);
+  if(type !== Ember.typeOf(b)) { return false; }
+  switch (type) {
+    case 'object':
+      return objectIsEquivalent(a, b);
+    case 'array':
+      return arrayIsEquivalent(a, b);
+    default:
+      return a === b;
   }
+}
 
+function arrayIsEquivalent(arrayA , arrayB) {
+  if(arrayA.length !== arrayB.length) { return false; }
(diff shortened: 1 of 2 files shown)


danielspaniel:
If you want to create tests, which I am glad you mentioned, all you have to do is make a unit test for this function and setup the conditions to test    
   obj1       obj2        expectedEqual 
  {e: [1]}    {e:[1]}        true
  {e:[1]}     {e:[1,2]}     false
  {e:[2,1]}  {e:[1,2]}     true ( but not so easy to make this work ) 

that sort of thing. 
 or is there something else you were thinking that I am missing?

bryanaka:
@danielspaniel Yes, I guess what I was asking for is what would be the preferred way to test this. To me, unit tests definitely make more sense than going through a full blown acceptance test, but just didn't know if there was a reason for not testing it in a unit test prior.

I've added unit tests now. Let me know if you would like to see more of anything :)

Also as far as code style goes, go ahead and comment on any pieces you would like me to change. The main piece is instead of `let varName = function() {...}` I've done `function fnName() {...}` to reduce the need for babel transpilation (I believe the output is almost equivalent to my code, just more obfuscated).

danielspaniel:
Very good point though, because I kind of blew off the tests ( which is odd for me ) .. since I was using the function as it was going through acceptance tests. But these tests are much better to have. So thanks mucho :)

bryanaka:
No problem! Thanks for your work on this project and glad I could help.

PR #192, opened 2016-04-21
Wrap arrays that use prototype extensions in Ember.A()

gilest:
Continuation of #114 #78


danielspaniel:
thanks @gilest, I guess I continually forget that people might not have prototype extensions turned on, probably because I am not sure why they would ever want to do that . .. anyway .. thanks for the touch ups.

gilest:
You're welcome @danielspaniel. I don't blame you – it's a real pain having them turned off.

In this case the project uses [dash.js](https://github.com/Dash-Industry-Forum/dash.js/) tech in a video player. The dash.js tech refuses to load with array extensions turned on.

PR #199, opened 2016-05-11
Add support for primaryKey in serializers

mattmcginnis:


danielspaniel:
that was quick .. thanks .. will start noodling

danielspaniel:
You had the right idea with your tests and I fully approve of what you did. The only problem was that this is very hard to test. 
 There is the concept of testing something on all the adapters ( that is where the shared-adapter-tests comes in ) .. where you do a certain data thing ( ask for data with build or make or mockFind ) and test the same thing on each adapter. 
 It is hard because your trying to pretend that all adapters/serializers are now in JSONAPI mode or RESTAPI mode. In the case of the acceptance tests, you are 100% in JSONAPI mode.  But what if you want to test REST API mode?  I figured the acceptance tests was not the place to test that ( though I am open to challenge ) .. but the place was the specific adapter/serializer pair tests ( JSONAPI, REST, ActiveModel )
 So, the answer is to set the application adapter/serializer to RESTAPI for cat view test ( by doing hokus pokus ) 
 or make a shared adapter /serializer test for that case of primaryKey setting ( by doing hokus pokus ) ..  
 Either way is fine but both involve a bit of whackiness.

mattmcginnis:
I could just add a test like this to each adapter test:

```javascript
test("adds a primaryKey attribute when serializer specifies the primaryKey", function () {

  let buildJson = build('cat');
  buildJson.unwrap();

  let expectedJson = {
    'cat': {
      id: 1,
      _id: 1,
      name: 'whatever',
    }
  };

  deepEqual(buildJson, expectedJson);
});

```

and then do the same for dog.  So, this tests that the primaryKey is there, which is sufficient, but does not manifest the actual error I encountered.  Let me know if this works for you.

mattmcginnis:
heh, actually that doesn't work anyway...

mattmcginnis:
It should be sufficient to just add a custom adapter for cat like:

```javascript
//tests/dummy/app/adapters/cat.js
import DS from 'ember-data';

export default DS.RESTAdapter.extend({
});
```
but this does not work for some reason.  Still investigating what hocus pocus is required.

mattmcginnis:
*punt* Any ideas?

danielspaniel:
Yes, I have refactored things around and will push it out tomorrow .. ( and will explain what I did before that .. you are on the right track by the way .. there was just one little thing you missed ( but I explain tomorrow )

PR #200, opened 2016-05-11
Support embedding records using `deserialize: 'records'`

patocallaghan:
Currently factory-guy only supports embedding records using:

```
attrs: {
  reviews: {embedded: 'always'}
}
```
but Ember Data also allows you to embed records using:

```
attrs: {
  reviews: {deserialize: 'records'}
}
```

From [the API docs](http://emberjs.com/api/data/classes/DS.EmbeddedRecordsMixin.html#toc_using-embedded-records)

> The attrs option for a resource `{ embedded: 'always' }` is shorthand for: `{ serialize: 'records', deserialize: 'records' }`


danielspaniel:
Good catch ..

PR #204, opened 2016-05-23
Fix handling of non-model-fragment hashes

mreq:
Regular hash objects didn't work before

    model-definition.js:50 Uncaught TypeError: Cannot read property 'match' of undefined

Example hash:

```coffee
FactoryGuy.define 'modelName',
  default:
    foo: { bar: true }
```


danielspaniel:
Thanks for the PR .. good catch .. and can add a test for this .. ??

mreq:
can do, will update once I get some time

iraspopo:
I run into this issue and was wondering when this fix will be available (merged in)?

danielspaniel:
I will work on this one today .. might be able to wrap it up .. stay tuned

danielspaniel:
Question .. what version of factory guy you on ..I just wrote a test for this and it works fine to use hash at attribute for non model fragment. 

 You will have to update factory guy ( and rerun that test ) 
  or send me your model, and your factory and your test that is failing so i can try and recreate it

danielspaniel:
Never mind .. I found the bug .. you have to pass in the hash in the build options .. ok .. I am all set

danielspaniel:
I have this fixed .. will release tomorrow

danielspaniel:
Fixed in new patch version:  
 https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.8.1

PR #211, opened 2016-06-16
Adding factories to the app instead

taras:
I don't know if this is the right solution but it seems like it could be. I tried to test it, but the demo app's index.html has all of it's code commented out. Why is that?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 var path = require('path');
+var fs = require('fs');
 var Funnel = require('broccoli-funnel');
 var mergeTrees = require('broccoli-merge-trees');
 var EmberApp = require('ember-cli/lib/broccoli/ember-app');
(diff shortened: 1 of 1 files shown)


danielspaniel:
You know that is a great question .. why is that index.html code commented out?  I have no idea?

taras:
It looks like you commented it out here https://github.com/danielspaniel/ember-data-factory-guy/commit/3a15f099ba5e8c712f74e95d89c4156d128c7a04

danielspaniel:
Just forgot what was going on there? I will fix that

danielspaniel:
please uncomment that if that helps you test it ..

danielspaniel:
I am testing your idea now

danielspaniel:
not working ..

danielspaniel:
wait .. take that back .. it is in the assets.[appname].js file .. 
 I think the only thing missing now is to require them because they don't get loaded unless they are required .. which is what factory guy manual setup does .. goes through the files and requires them 
 so .. wow .. awesome dude .. I think you did that in 10 minutes ? .. amazing

taras:
:) I'm glad it did the trick. Do you need anything else?

danielspaniel:
yes, can you explain about the microstates sometime to me .. me want to understand more about that .. I don't quite get it

taras:
Let's do a little screenshare and I'll show you how it works.

danielspaniel:
let me know when you want to do that

PR #212, opened 2016-06-23
Added failing test for `returns` with mockCreate

patocallaghan:
I'm seeing some weird behaviour with `returns` not working but not sure if it's because I'm doing something wrong.

Essentially in the acceptance test I do:

``` js
let project = make('project'); //project record has id 1
mockCreate('project').match({title: newProjectTitle}).returns({model: project})
```

yet in the application code that calls `createRecord` the record is a newly created one and not the one from the `returns`

``` js
this.get('store').createRecord('project', {title: title, user: user}).save()
.then((record) => {
  // project record has id 2
});
```

The consequence of this is that the `record.reload()` fails due to `mockReload` being passed the wrong record id in the test.

Am I doing something wrong here or is this a bug?

### DIFF
--- a/tests/acceptance/user-view-test.js
+++ b/tests/acceptance/user-view-test.js
@@ -58,10 +58,18 @@ test("Add a project to a user with mockCreate", function () {
 
     // Remember, this is for handling an exact match, if you did not care about
     // matching attributes, you could just do: TestHelper.mockCreate('project')
+
+    // USING BUILD
+    // let project = build('project');
+    // mockCreate('project').match({title: newProjectTitle}).returns({project});
+    // mockCreate('project').match({title: newProjectTitle}).returns({project: project});
+
+    // USING MAKE
     let project = make('project');
+    mockCreate('project').match({title: newProjectTitle}).returns({project: project.id});
+
     mockReload('project', project.get('id'));
     console.log('Project record ID returned in Test code:', project.get('id'));
-    mockCreate('project').match({title: newProjectTitle}).returns({model: project});
 
     /**
      Let's say that clicking this 'button:contains(Add New User)', triggers action in the
--- a/tests/acceptance/user-view-test.js
+++ b/tests/acceptance/user-view-test.js
@@ -1,4 +1,4 @@
-import { make, build, buildList, mockFind, mockCreate } from 'ember-data-factory-guy';
+import { make, build, buildList, mockFind, mockCreate, mockReload } from 'ember-data-factory-guy';
 import moduleForAcceptance from '../helpers/module-for-acceptance';
 
 moduleForAcceptance('Acceptance | User View');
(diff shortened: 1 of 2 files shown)


danielspaniel:
Yeah .. this is a very tricky example of the returns being different got GET requests and the create POST and PUT, update. 

 In the docs for mockCreate for example I have this:

```js
 // Returning belongsTo relationship. Assume outfit belongsTo 'person'
  let person = build('super-hero'); // it's polymorphic
  mockCreate('outfit').returns({ person });
```

Do you see the difference? 

I knew this would be a problem one day, but I am not sure how to fix or make this clearer or make it different than the GET style? Any ideas.

patocallaghan:
I *think* I understand.

Are you saying that the API for `mockCreate` doesn't support the `.returns({model: user})` or `.returns({json: user})` syntax? Only the GET requests like `mockQuery` and `findRecord` support it?

danielspaniel:
ok .. almost .. and your pointing to the fact that I have to make the docs way clearer ( which I will ) 

So, you did this:

```js
let project = make('project'); //project record has id 1
mockCreate('project').match({title: newProjectTitle}).returns({model: project})
```

and you need to do this;

```js
let project = build('project'); //project record has id 1
mockCreate('project').match({title: newProjectTitle}).returns({project})
```
or to use existing model 
```js
let project = make('project'); //project record has id 1
mockCreate('project').match({title: newProjectTitle}).returns({project: project.id})
```

so in mockUpdate, mockCreate  the returns does not use the {json: json} or {model: model} concept but you are literally passing the attributes you want to be returned

danielspaniel:
the reason the returns is not the same here as returns for GET requests where you passing in ONE payload of model/modles or json is:
  here you are saying .. hey, mock that request and possibly match these arguments ( parameters ) .. and then possibly pass back a few extra parameters here .. so the returns is just a list of parameters to add to the response and not the response itself ( if you know what I mean )

patocallaghan:
Okay cool, I understand. Saying that, I've updated my failing test to use each of your examples above but still can't get the test to pass. The app code still always seems to use a new record rather than the one I define in the test.

You can test each scenario by commenting in the relevant lines. The main issue I'm having is because the record used in the app code is different than the record in the test my `mockReload` is failing. 

```
// USING BUILD
// let project = build('project');
// mockCreate('project').match({title: newProjectTitle}).returns({project});
// mockCreate('project').match({title: newProjectTitle}).returns({project: project});

// USING MAKE
let project = make('project');
mockCreate('project').match({title: newProjectTitle}).returns({project: project.id});
```

danielspaniel:
Ok .. I think I see what is going on. 

Your saying mockCreate('project').returns( the project I want to return that I already made )

so, I was wrong in telling you what to do because it was late, I was confused too and ?? 

Anyway .. here is what will work for you:

let project = make('project')
mockCreate('project').match({title: newProjectTitle}).returns({id: project.id});

Your saying, hey .. you're about to make a project, and return this attribute ( id ) instead of making a new one with new id, and here is the id I want to use ( project.id ) because I already made a new one

patocallaghan:
OK, that did the trick!! Thanks 🎉 Closing this now.

Issue #213, opened 2016-06-29
match() fails with JSONAPI for relationship when serializer's payloadKeyFromModelName() doesn't pluralize


cristinawithout:
Our application serializer has the following method that purposely does not pluralize.

`payloadKeyFromModelName(modelName) { return Ember.String.underscore(modelName); }`

mock-create-request.js is [setting serializeMode to true ](https://github.com/danielspaniel/ember-data-factory-guy/blob/v2.7.0-beta.7/addon/mocks/mock-create-request.js#L67) and jsonapi-fixture-converter.js [uses that to pluralize the type](https://github.com/danielspaniel/ember-data-factory-guy/blob/v2.7.0-beta.7/addon/converter/jsonapi-fixture-converter.js#L15).

So when I try to use mockCreate().match() with a relationship, the type key is getting pluralized in the expected data and the match fails because our serializer does not pluralize the type key.

Example:
`mockCreate('entry-type').match({name: 'My Name', handle: 'my_name', section: {id: '1', type: 'section'}})`

And when it reaches [this comparison](https://github.com/danielspaniel/ember-data-factory-guy/blob/v2.7.0-beta.7/addon/mocks/mock-create-request.js#L74), the data is:

expectedData.data:
```
{
attributes: {
	handle: "my_name",
	name: "My Name"
},
relationships: {
	section: {
		data: {
			id: "1",
			type: "sections" // pluralized
		}
	}
},
type: "entry-types" // pluralized
}
```

requestData.data:

```
{
attributes: {
	handle: "my_name",
	name: "My Name"
},
relationships: {
	section: {
		data: {
			id: "1",
			type: "section" // not pluralized
		}
	}
},
type: "entry-type" // not pluralized
}
```

So the match fails, but the requestData is exactly as it should be using our serializer.

Is it possible to instead lookup serializers by modelName and use payloadKeyFromModelName() as the JSONAPI serializer does when serializing the type?

danielspaniel:
@cristinawithout .. good point here. 
 Do you think you could put together a PR that would handle this ?? 
 and put a nice test together ( which would be pretty similar to what you have above ) ??

danielspaniel:
v2.7.0-beta.8 has your changes

PR #214, opened 2016-06-29
Issue 213 jsonapi payloadkey

cristinawithout:
For #213


danielspaniel:
thanks .. @cristinawithout  .. 👍

danielspaniel:
I think this is applicable to the rest converter too .. don't you think?

danielspaniel:
I can hook that up and make a test for it .. so not asking you to do that.

cristinawithout:
Sorry. Missed your replies on this one. I was thinking it might apply to other converters too, but I wasn't sure which.

danielspaniel:
not a big deal, since it is a mixin .. so if anyone peeps up I can add it pretty quickly.
 but I can't think of why you would need for a  delete ? ( because find / query makes no sense )

cristinawithout:
Nah, I can't think of a reason the other types would need match(). Was talking about this older type transform by payload key stuff.

cristinawithout:
You probably already took care of that though.

danielspaniel:
I made it so that payload key was used by all adapters if they had it not just JSONAPI .. maybe it was already like that but I made the tests prove it ( I think ? ) you had just the one test in the JSONAPI adapter custom

Issue #216, opened 2016-07-06
.match() for mockUpdate()


cristinawithout:
I don't see a way to verify the data sent to my endpoint on update is correct as I can with mockCreate().match()

danielspaniel:
err .. um .. well .. let's see, I am digging through my thoughts for this one, and the answer is?
 no found 
oh well .. 
 not sure why I never did that .. is that something nice to have .. 
 I just never thought of it 
 since I usually set stuff and update it and only care about returning extra attrs but never checking that the attrs I am sending are this way or that 
 
 but you know, it's not a bad idea 

 might be easy to do 
  
 you want to take a shot at it?
 maybe give me example of what you want to do so I can be 100% certain I don't mess it up.

cristinawithout:
I'll see if I can make a PR with at least some tests.

danielspaniel:
v2.7.0-beta.10 has this ability

danielspaniel:
@cristinawithout =>  side note. I thought this was a pretty silly feature at first, and lo and behold as soon as I finished it, I used it right away 👍 

seems to happen every time I get a feature request. I kind of think ( who would ever need that ). Then it ends up being me using it pretty much the next day.

cristinawithout:
👍  Nice. I'm using it along with mockJack's throwUnmocked:true in acceptance tests to make sure A. all the form inputs are hooked up to the buffered model (our own implementation similar to buffered proxy - but handles relationships) B. the values in the buffered model transfer to the model on save  C. The values are serializing as expected.

Using match(), I can make sure all of those things are working. Without it, the update would return success, and the test would pass even if the the model wasn't connected to the inputs and/or the serializer was returning nothing.

An update acceptance test where I'm using it is basically like this:

```
test('can update entry type', function(assert) {
    mockFindAll('section', 5);
    mockFindAll('entry-type', [{
        id: '100',
        name: 'My Entry Type',
        handle: 'my_entry_type',
        section: {
            id: '5',
            type: 'section',
        }
    }]);
    let mock = mockUpdate('entry-type', 100)
        .match({
            name: 'New Name',
            handle: 'new_name',
            section: {id: '1', type: 'section'}
        });

    visit('/entry-types/100/update');

    // Edit the name, handle and section relationship.
    fillIn('[data-test=et-name] input', 'New Name');
    fillIn('[data-test=et-handle] input', 'new_name');
[...]

PR #217, opened 2016-07-06
Tests for mockUpdate.match #216

cristinawithout:
Here's some tests for mockUpdate().match().

The tests pass and it works how I need in my own app's tests when I copy the methods match(), extraRequestMatches(), and attributesMatch() from mock-create-request.js to mock-update-request.js.

But you probably want something cleaner than a copy/paste job. Not sure if there would be any variation  for update. If not, maybe a 'mock-match-request' class with those methods which both create and update extend.

### DIFF
--- a/addon/mocks/mock-create-request.js
+++ b/addon/mocks/mock-create-request.js
@@ -1,24 +1,17 @@
 import Ember from 'ember';
 import FactoryGuy from '../factory-guy';
-import MockRequest from './mock-request';
-import {isEquivalent} from '../utils/helper-functions';
-const { isPresent, isEmpty } = Ember;
+import MockRequest from './mock-request-match';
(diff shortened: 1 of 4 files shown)


cristinawithout:
Went ahead and committed the update via a 'mock-request-match' class that both update and create inherit from. Feel free to use or go with a different approach.

danielspaniel:
I am thinking that a mixin would be nicer .. they are hard to do with classes .. but I will give it a whirl
 I am shocked that this was so easy .. how nice though. 

  Wish everything was that simple.

danielspaniel:
All set .. converted to mixin style.
 Will ship this out tomorrow

danielspaniel:
v2.7.0-beta.10 has this new code .. thanks for the help 👍

PR #221, opened 2016-07-26
mockFind does not return a failed response

daibhin:
```
not ok 368 PhantomJS 2.1 - DS.RESTAdapter/JSONSerializer | #mockFind: failure with fails method
    ---
        actual: >
            false
        expected: >
            true
        stack: >
            http://localhost:7357/assets/test-support.js:4147:17
            exception@http://localhost:7357/assets/vendor.js:59253:9
            onerrorDefault@http://localhost:7357/assets/vendor.js:51230:33
            trigger@http://localhost:7357/assets/vendor.js:73618:19
            http://localhost:7357/assets/vendor.js:74869:40
            invoke@http://localhost:7357/assets/vendor.js:19475:20
            flush@http://localhost:7357/assets/vendor.js:19539:17
            flush@http://localhost:7357/assets/vendor.js:19347:22
            end@http://localhost:7357/assets/vendor.js:19702:30
            run@http://localhost:7357/assets/vendor.js:19824:21
            join@http://localhost:7357/assets/vendor.js:19844:30
            join@http://localhost:7357/assets/vendor.js:40323:33
            error@http://localhost:7357/assets/vendor.js:102546:39
            fire@http://localhost:7357/assets/vendor.js:7906:36
            fireWith@http://localhost:7357/assets/vendor.js:8036:11
            done@http://localhost:7357/assets/vendor.js:13478:24
            http://localhost:7357/assets/vendor.js:13842:17
        message: >
            Error: Ember Data Request GET /profiles/1 returned a 404
            Payload (text/html; charset=utf-8)
            Not found: /profiles/1/tests/index.html?hidepassed
        Log: |
    ...
```

### DIFF
--- a/tests/unit/shared-factory-guy-test-helper-tests.js
+++ b/tests/unit/shared-factory-guy-test-helper-tests.js
@@ -43,7 +43,7 @@ SharedBehavior.mockFindCommonTests = function() {
 
       equal(mock.get('catName'), 'Cat 1');
       equal(mock.get('catFriend'), 'Friend 1');
-      
+
       FactoryGuy.store.find('cat', mock.get('id')).then(function(cat) {
         equal(cat.get('name'), 'Cat 1');
         equal(cat.get('friend'), 'Friend 1');
@@ -118,6 +118,24 @@ SharedBehavior.mockFindCommonTests = function() {
         ok(profile.get('description') === 'dude');
         done();
       });
+    });
+  });
+
+  test("failure with fails method", function(assert) {
+    Ember.run(()=> {
+      let done = assert.async();
+
+      let profile = make('profile');
+      let profileId = profile.get('id');
+      let mock = mockFind('profile').fails();
+
+      FactoryGuy.store.findRecord('profile', profileId)
+        .catch(()=> {
+            equal(mock.timesCalled, 1);
+            ok(true);
+            done();
+          }
+        );
     });
   });
 


danielspaniel:
are you adding this test to show that the fails method works?

daibhin:
@danielspaniel was using factory-guy today and trying to `mockFind` and return a failed request. Couldn't get it to work locally so I thought I'd see could I reproduce it in the test suite.
(I didn't actually mean to submit a PR)

Is this a bug with factory-guy that `mockFind` isn't returning a failed response?

danielspaniel:
You kind of did something unusual .. but you know, this is a pretty whopper bug ( I think ) .. so thanks for bringing it up.
 
 I think the API is a bit confusing too .. in that you have to do mockFind('profile', {id: id}) to get that to match, but even then there are troubles in testing .. but I am working on it. 

Bottom line, your test brought up a good issue, so thanks

danielspaniel:
Ok .. what is happening is:

 you make profile .. it is now in the store. 
 you call findRecord, it returns immediately the record, while it does the network call in the background ( default behaviour ) 
 so the done() call was never reached in the catch statement since the catch was not handled, since the then was already handled by the returned record
 
 long story, but something that would not happen if you did 

  mockFind without the make first 

 still, the way this works is sucky. 

 in that you should not do mockFind('profile') .. that was root of issue. 

  should be mockFind(model)  or mockFind(modelName, id) 
 
 what you have to do now is mockFind(modelName, {id: model.id}) 
 
 things changed a while ago and your test now will make sure that it is fixed and I improve that a bit 
So, try that mockFind('profile', {id: profile.id}) 
 and you will find that this works ( even though your route will still have to handle the errror )

danielspaniel:
so, I just fixed it so you can now do this:

```js
     let profile = make('profile');
     let mock = mockFind(profile).fails();
     FactoryGuy.store.findRecord('profile',  profile.id));
```

so, this makes it smooth and easy to mock the find when you made the model 
 stay tuned for v2.7.2 out tomorrow.

daibhin:
Awesome! Thanks @danielspaniel 😀

danielspaniel:
v2.7.2 is now ready .. and your welcome @daibhin .. great test by the way .. blew a hole is factory guy with one simple shot.

PR #229, opened 2016-08-05
Update README to reflect mockTeardown() usage in component tests

oniofchaos:
Addresses answer to https://github.com/danielspaniel/ember-data-factory-guy/issues/224

I just noticed my Vim settings stripped out all the extra whitespace in the README. Let me know if you want me to revert those portions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1024,7 +1024,8 @@ test("Using FactoryGuy.cacheOnlyMode with except", function() {
 ### Testing models, controllers, components
 
 - FactoryGuy needs to setup the factories before the test run.
-  - use manualSetup and mockTeardown functions to set up FactoryGuy in unit/component tests
+  - By default, you only need to call `manualSetup(this.container)` in unit/component tests
+  - If you are using mock methods, such as mockQuery, use `mockSetup()` and `mockTeardown()`
 
 - [Sample model test (profile-test.js):](https://github.com/danielspaniel/ember-data-factory-guy/blob/master/tests/unit/models/profile-test.js)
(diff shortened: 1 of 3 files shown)


danielspaniel:
Ooops. I think I mistakenly gave the impression that I forgot to put the afterEach() mockTeardown in the moduleForComponent and moduleForModel tests. 

What I meant was .. 

 oops .. I forgot to tell people that if they are going to use mocking in component and model tests ( rare ..but I do it myself ) .. that they need to add mockSetup, mockTeardown. 
 
 Sorry about that .. I think all you have to do is add to the readme and maybe setup up a test if you want that uses the mocks ( component or model ) ..

 but it has to be clear that if you just make or makeList or build/ buildList .. you don't need mockSetup/mockTeardown 

 know what I mean?

oniofchaos:
I'm not sure what you mean, actually :)

If I'm using `make` in a component test and don't include `manualSetup(this.container)`, the test fails. 

Are you suggesting that `mockSetup()` should be used instead of `manualSetup(this.container)`?

danielspaniel:
No, 
 what I mean is:
 
 for moduleForComponent or moduleForModel tests
 you only need need need manualSetup(this.container)

 and by any chance you are doing mock this or that then you 
 need the mockSetup, mockTeardown

 your documentation is showing you ALWAYS need mockSetup, mockTeardown.

 see what I mean?

oniofchaos:
Ok, check that out.

danielspaniel:
sold .. thanks for listening and getting this right. 👍

PR #243, opened 2016-09-07
Failing test for having a type "property" on your model

patocallaghan:
This is a failing test for Issue https://github.com/danielspaniel/ember-data-factory-guy/issues/242

Running the `ember test --server --filter="default employee"` will show the failing tests


patocallaghan:
Do you want to keep this test or do you have one already? Adding the `polymorphic: false` prop causes the test to pass now

danielspaniel:
I have a test already ( on the cat model ) so I don't need this one

patocallaghan:
Cool, no problems. Sorry for the false alarm 😄

PR #251, opened 2016-09-19
Use Array.includes instead of deprecated contains if possible

mdentremont:
- Ember 2.8 deprecated Ember.Array.contains
- Use Ember.array.includes if available

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -522,7 +522,8 @@ class FactoryGuy {
       let adapter = findAdapter(name);
       let shouldCache = ()=> {
         if (Ember.isPresent(except)) {
-          return (Ember.A(except).contains(name));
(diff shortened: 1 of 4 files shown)


danielspaniel:
Do you think we can get away with outright replacing contains with includes. This works on ember 2.5 and I think babel does polyfill for us? Not sure? Was there reason you did conditionally use includes or contains ??

mdentremont:
I included contains just to be safe as I'm not very familiar with how backwards support works for addons.

danielspaniel:
Me too .. I think it might have to do with babel actually .. but not sure .. checking

danielspaniel:
Just looked it up .. looks like Babel polyfills so we don't have to worry about it.

danielspaniel:
I am going to merge this just in case though

PR #252, opened 2016-10-13
Fix a link to mockFindRecord in the readme

whatthewhat:
Clicking the mockFindRecord link will now take you to the mockFindRecord section in the readme, which is probably what you expect to happen :)


danielspaniel:
Thanks Mikhail 😄

PR #253, opened 2016-10-17
addIncludedArray() replacing payload data with included

cristinawithout:
I have a model 'department', which is hierarchal. Each department can have a parent. So the department model has a relationship with its own model type 'department'.

In my tests, I have a trait called 'parent' and one called 'primary'.

```
FactoryGuy.define('department', {
    traits: {
        primary: {
            name: 'Primary Department',
            parent: FactoryGuy.belongsTo('department', 'parent'),
        },
        parent: {
            name: 'Parent Department',
            parent: null
        }
    }
//...
});
```

When I use `buildList()` and include the 'primary' trait, the result I get back contains only the 'parent' model.

Stepping through, I found this is because addIncludedArray() in rest-fixture-converter is setting `payload[key] = this.included[key];`. Since the key 'department' is present in 'included', it replaces everything that in the payload when I want it to add it to the payload.

I wasn't sure where the best place to add this to tests would be, but I can put one together if you let me know.

### DIFF
--- a/addon/converter/rest-fixture-converter.js
+++ b/addon/converter/rest-fixture-converter.js
@@ -53,7 +53,11 @@ export default class extends JSONFixtureConverter {
    */
   addIncludedArray(payload) {
     Object.keys(this.included).forEach((key)=> {
-      payload[key] = this.included[key];
+      if (!payload[key]) {
+        payload[key] = this.included[key];
+      } else {
+      	Array.prototype.push.apply(payload[key], this.included[key]);
+      }
     });
   }
 


danielspaniel:
Mmm ... let's see.. seems like this ( if it only applies to rest adapter ( and I think it does ) .. would go in the rest adapter custom buildList tests ( I know there is custom build so if there is no custom buildList then make that module ) 
 I would greatly appreciate the tests too so thanks for asking about that

cristinawithout:
Thanks. I'll try to get to the test within the next few days.

danielspaniel:
@cristinawithout .. if you want some help testing this let me know

cristinawithout:
Sorry. I haven't had time to get to it yet. It should be a fairly easy test. If you want to take care of it, I wouldn't mind at all.

danielspaniel:
fixed in =>  https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.9.3

PR #255, opened 2016-10-25
Add support for using factory-guy in other environments

juwara0:
Closes: https://github.com/danielspaniel/ember-data-factory-guy/issues/254

CHANGELOG:
**Updated** index.js to support using factory-guy in other environments
**Updated** README to show configuration needed for using factory-guy in other environments

DESCRIPTION:
This PR adds support for using factory-guy in other environments to the base cases of `development` and `test`.

The user just needs to add:

```
// file: config/environment.js
ENV.factoryGuy = true;
``` 

to the environments in which factory-guy is desired.

By default, the test environment is still enabled but now a user can turn it off by specifying:

```
// file: config/environment.js
if (environment === 'test') {
  ENV.factoryGuy = false;
  ...
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -812,6 +812,10 @@ You would use this to make models like:
 
 - You can set up scenarios for you app that use all your factories from tests
 - In config/environment.js place a flag => factoryGuy: true
+- NOTE: Do not set the flag => factoryGuy: true in the `test` environment. Factories are enabled
+  by default for the `test` environment and setting the flag tells factory-guy to load the app/scenarios
+  files which are not needed for using factory-guy in testing. This will result in errors being generated if
+  the app/scenarios files do not exist.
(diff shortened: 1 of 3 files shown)


danielspaniel:
Looks good Michael
I was just about to ask how this was going. Ok .. one question. 
 The only caveat is to make sure in test environment to not  put =>  factoryGuy:true .. or else it would look for the scenarios/main.js file and barf if not found.

juwara0:
@danielspaniel - I have added a note to the README to highlight that `factoryGuy: true` should not be added to the `test` environment.

danielspaniel:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.8.2  has this PR included

PR #257, opened 2016-11-03
Fix URLs for query and queryRecord

Asherlc:
Currently, the URLs generated from mockQuery and mockQueryRecord don't singularize/pluralize appropriately, and they don't include the query hash that's passed in. This fixes that, so the requests are actually mocked.

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -503,9 +503,9 @@ class FactoryGuy {
    @param {String} id
    @return {String} url
    */
-  buildURL(modelName, id = null, requestType) {
+  buildURL(modelName, id = null, requestType, query) {
     const adapter = this.store.adapterFor(modelName);
-    return adapter.buildURL(modelName, id, null, requestType);
+    return adapter.buildURL(modelName, id, null, requestType, query);
   }
 
   /**
--- a/addon/mocks/mock-query-record-request.js
+++ b/addon/mocks/mock-query-record-request.js
@@ -14,4 +14,8 @@ export default class MockQueryRecordRequest extends MockQueryRequest {
     this.setResponseJson(FactoryGuy.fixtureBuilder.convertForBuild(modelName, null));
     this.setValidReturnsKeys(['model','json','id','headers']);
   }
+
+  getUrl() {
+    return FactoryGuy.buildURL(this.modelName, null, 'queryRecord', this.get('query'));
+  }
 }
--- a/addon/mocks/mock-query-record-request.js
+++ b/addon/mocks/mock-query-record-request.js
@@ -16,6 +16,6 @@ export default class MockQueryRecordRequest extends MockQueryRequest {
   }
 
   getUrl() {
-    return FactoryGuy.buildURL(this.modelName, null, 'queryRecord', this.get('query'));
+    return FactoryGuy.buildURL(this.modelName, null, 'queryRecord', this.queryParams);
   }
 }
--- a/addon/mocks/mock-query-request.js
+++ b/addon/mocks/mock-query-request.js
@@ -10,4 +10,7 @@ export default class MockQueryRequest extends MockGetRequest {
     this.queryParams = queryParams;
   }
 
+  getUrl() {
+    return FactoryGuy.buildURL(this.modelName, null, 'query', this.get('query'));
+  }
 }
--- a/addon/mocks/mock-query-request.js
+++ b/addon/mocks/mock-query-request.js
@@ -11,6 +11,6 @@ export default class MockQueryRequest extends MockGetRequest {
   }
 
   getUrl() {
-    return FactoryGuy.buildURL(this.modelName, null, 'query', this.get('query'));
+    return FactoryGuy.buildURL(this.modelName, null, 'query', this.queryParams);
   }
 }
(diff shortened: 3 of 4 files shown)


danielspaniel:
@Asherlc , I welcome this fix, but can you make it so build is not broken ??  maybe add a few tests here and there .. ??

Asherlc:
Hi, happy to write some tests. As far as the build failing, I'm a bit confused about why that's happening (it works for me locally). I'm not super familiar with Travis CI -- is there a way to see the source of the error?

danielspaniel:
I the merge box where is says:

 "continuous-integration/travis-ci/pr" -- The Travis Cl....      Details 

 click on Details button, and then when you see the build you see the various failed builds .. click on one of those .. and voila .. you will be here:  https://travis-ci.org/danielspaniel/ember-data-factory-guy/jobs/173095304

Asherlc:
Is it expected to see 404 errors in the console when running the test suite (on master)? My suspicion is that this might be the source of the tests passing -- that mockjax doesn't intercept the request, but nothing actually throws an error if a 404 is hit.

screen shot 2016-11-04 at 3 44 57 pm

danielspaniel:
Yes, those are supposed to "not be caught" by FG

like this test:

```javascript
  test("fails when match args not present in createRecord attributes", function(assert) {
    Ember.run(()=> {
      let done = assert.async();

      let mock = mockCreate('profile').match({ description: 'correct description' });

      FactoryGuy.store.createRecord('profile', { description: 'wrong description' }).save()
        .catch(()=> {
          ok(true);
          equal(mock.timesCalled, 0);

          done();
        });
    });
  });
```

will throw error out to console because the mock does not match .. and it is not supposed to match. 
So, it is a bit tricky to write tests that say "here is the mock testup", and you call this query .. the mock will not be handled .. 
maybe there is better way ?  I am open to better idea

Asherlc:
Makes total sense. Would you expect `Ember.$.mockjax.mockedAjaxCalls()` to include calls from `FactoryGuy.store.queryRecord`? I'm trying to figure out how to verify that the actual ajax request that I expect to happen in the test actually was requested (I don't see it in the actual ajax request log).

danielspaniel:
Ember.$.mockjax.mockedAjaxCalls() will tell you which were handled ( which mockjax returned ) .. but not sure if this well help .. as you say .. you want unhandled requests.

 How about if you just test the getUrl method on MockCreateRequest?  
setup a mock and say .getUrl()  and see if it looks ok ?
   
   It is not the same as the query with params matching the mock .. but might be good shortcut to same idea

Asherlc:
Ach, turns out this was all a red herring. Turns out we override the `urlForQueryRecord` to singularize it in our app, but it looks like `FactoryGuy.store` is built off of the built-in adapters (correct me if I'm wrong). So I guess my question changes... is there a way to make the mock* methods use the application's adapter, not the JSONAPI one?

danielspaniel:
Huh? Me confused .. 
 what you mean by "is there a way to make the mock* methods use the application's adapter, not the JSONAPI one?"
  what is => "the JSONAPI one"  ?

factory guy does this:

```javascript 
  /**
   Build url's for the mockjax calls. Proxy to the adapters buildURL method.

   @param {String} typeName model type name like 'user' for User model
   @param {String} id
   @return {String} url
   */
  buildURL(modelName, id = null, requestType) {
    const adapter = this.store.adapterFor(modelName);
    return adapter.buildURL(modelName, id, null, requestType);
  }
```

so it just asks the adapter what url should be so if you override that urlForQueryRecord this method ( FactoryGuy.buildUrl )  will return the correct url ( UNLESS ) .. you don't pass in the requestType 
 
 Right now I only am passing in the correct requestType for findRecord


```javascript
export default class MockFindRecordRequest extends MockGetRequest {

  constructor(modelName) {
    super(modelName);
    this.setValidReturnsKeys(['model', 'json', 'id', 'headers']);
  }

  getUrl() {
    return FactoryGuy.buildURL(this.modelName, this.get('id'), 'findRecord');
  }
}
```

but I can do the query/queryRecord ones too ( I had that as an issue anyway ) if you want are too confused ( like I am ) 

 just would be nice to get your issue nailed down.

Asherlc:
Ok, I think I get it now.

My issue is basically just https://github.com/danielspaniel/ember-data-factory-guy/issues/246. The tests pass in the factory-guy test suite because the urls for findRecord (without an ID) and queryRecord happen to be the same with the default ED adaptors. But when I ran it in my app, which overrides the queryRecord URL, they no longer matched up, and things broke.

Nothing to do with specifying an adaptor or anything like that, that was incorrect on my part.

So, yeah, my issue is just https://github.com/danielspaniel/ember-data-factory-guy/issues/246, if I finally have my head around it

danielspaniel:
Ok .. that was a good one ( we both running around like chickens ) .. oh well .. now I have good reason to fix that sucker.  Should be ready by end of weekend. Stay tuned.

danielspaniel:
Finally got this one @Asherlc , and did all the urlFor methods while I was at it: 
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.9.2

Asherlc:
Thanks! :)

PR #258, opened 2016-11-07
Fix typo for defining fragment factories

supremebeing7:


danielspaniel:
Yikes, how did I miss that .. thanks 👍

PR #261, opened 2016-11-29
Fix `Cannot read property '_super' of undefined` with Ember 2.10.0

GavinJoyce:
This should fix https://github.com/danielspaniel/ember-data-factory-guy/issues/260

TODO:

 * [x] Verify that this does indeed fix the issue
 * [ ] Write a test
 * [x] Figure out why this worked in `Ember 2.9` and not `Ember 2.10`

### DIFF
--- a/addon/converter/fixture-converter.js
+++ b/addon/converter/fixture-converter.js
@@ -15,7 +15,7 @@ import Ember from 'ember';
  If there are associations in the base fixture, they will be added to the
  new fixture as 'side loaded' elements, even if they are another json payload
  built whith the build/buildList methods.
- 
+
  @param {DS.Store} store
  @param {Object} options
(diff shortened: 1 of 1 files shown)


danielspaniel:
I am trying to think of how to test it as well ..

GavinJoyce:
I'm looking into `Figure out why this worked in Ember 2.9 and not Ember 2.10` first just in case there is an issue with the `2.10` release

danielspaniel:
sorry .. forgot to pull this .. i just changed the method and wrote the test ... but it's fixed now.
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.9.4

GavinJoyce:
Thanks!

PR #262, opened 2016-11-29
Remove some rogue console.logs

patocallaghan:
Looks like these were committed with the last commit


danielspaniel:
thanks again for the PR .. i just zapped it 2 mins ago .. check v2.9.5

PR #263, opened 2016-12-12
Allow mockCreate/mockUpdate to receive a matching function

lorcan:
**tl;dr** Allow `match` to accept a function or attribute(s) to determine whether a create or update matched.

The function passed to `match` is given the `requestData` and can perform bespoke logic to determine whether a request matched or not. If the function returns `true` then a match is recorded, otherwise it is not.

We have a use-case where we're updating a complex, nested model called `Parent`. `Parent` hasMany `Child` relationships, each of which has multiple attributes. We're mocking an update to `Parent` but what we really care about is one change to an attribute of `Child`, let's say `description`. Using the existing `match` function we would need to pass in the entire child structure, which is quite unwieldy. By allowing `match` to receive a function the responsibility of determining whether a match was made belongs to the test, e.g.,:

```
      let done = assert.async();
      let customDescription = "special description";
      let parent = make('parent');

      let updateMock = mockUpdate(parent);
     
      updateMock.match(function(requestData) {
        return requestData.children[0].description === customDescription;
      });
      profile.set('description', customDescription);
      profile.save().then(function(parent) {
        equal(updateMock.timesCalled, 1);
        done();
      });
```

One thing I don't like about this approach is that the caller needs to be aware of the adaptor type and process the `requestData` themselves to do the match. I think it's a reasonable compromise to get the flexibility of performing arbitrary matching.

I've given a stab at writing documentation and tests for this functionality. Happy to start a discussion and rework this PR if it seems like a sensible addition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1483,12 +1483,14 @@ Usage:
 ##### mockCreate
 
   - Use chainable methods to build the response
-    - match
-      - Attributes that must be in request json
+    - match: takes a hash with attributes or a matching function
+      1. attributes that must be in request json
         - These will be added to the response json automatically, so
           you don't need to include them in the returns hash.
         - If you match on a belongsTo association, you don't have to include that in
-          the returns hash either ( same idea ).
+          the returns hash either ( same idea )
+      1. a function that can be used to perform an arbitrary match against the request
+          json, returning `true` if there is a match, `false` otherwise.
     - returns
       - Attributes ( including relationships ) to include in response json
   - Need to wrap tests using mockCreate with: Ember.run(function() { 'your test' })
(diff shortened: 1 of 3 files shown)


danielspaniel:
I guess this is not too bad an idea, and you pointed out the right flaws. But if you really need to get this down deep, there is no choice.  
 I can't right now think of another way to do this without the function. 

 And I really like / am shocked that you lay out the tests quite thoroughly and update the documentation. This almost never happens, so I just got a bruise when I fell out of my chair.

lorcan:
Lol! Awesome! :-D
We're huge fans of ember-data-factory-guy!

PR #264, opened 2016-12-12
Should matching non-existent attributes fail?

lorcan:
**tl;dr** Passing an attribute to `match` with a bad/unknown name passes silently.

If you have a typo in attribute name  passed to `match` the test will pass, which is a little surprising. I would have expected this to fail. Feels a little dangerous to silently pass this case.

Happy to flesh this one out if you have any pointers. A naive solution might be to count the keys in the hash passed to `match` and ensure we're checking that many attributes in `matchCheckKeys` in `AttributeMatcher`?

### DIFF
--- a/tests/unit/shared-factory-guy-test-helper-tests.js
+++ b/tests/unit/shared-factory-guy-test-helper-tests.js
@@ -1139,6 +1139,23 @@ SharedBehavior.mockCreateTests = function() {
     });
   });
 
+  test("don't match bogus attributes", function(assert) {
+    assert.expect(2);
+    Ember.run(() => {
+      let done = assert.async();
+      let customDescription = "special description";
+
+      let mock = mockCreate('profile').match({bogus: customDescription});
+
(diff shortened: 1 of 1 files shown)


danielspaniel:
I guess I figured that if you declare an attr that should match and it is not an attribute then we ignore it.  Could be many reasons for being bogus, but you make an interesting point.  Why not let the user know.  I think the best thing that I can think of right now is we issue a warning, but don't try and change how we match.  
 So in this case you would see the warning "That bogus attribute is not found on that model"  
 And that way they can see the warning in the console and fix it rather than us trying to be clever.

 What you think about that plan?

lorcan:
I think this makes perfect sense! I don't think we should be any smarter in the matching. 

How would you feel about an assertion check? Too much?

danielspaniel:
You know. I think you have something here, because if you are going to put something in match you expect it to do something, so therefore an assertion would make sense.  
 I use warnings to tell you if your trait is not found, but now that I think of it, an assertion would be better there as well. 
 So .. I second the motion on assertion.

PR #265, opened 2016-12-12
Misc. typo fixes

lorcan:
Came across a couple of typos in the Readme and decided to run a spell check. Hope you don't mind and I'm not being too pedantic ;-)


danielspaniel:
No at all .. pedantic is good in these cases, because it makes everything more sharp. Especially in documentation, things should be as close to perfect as possible. Thanks for the efforts.

PR #266, opened 2016-12-15
Use assert rather than warn for missing traits

lorcan:
We kinda discussed using assertions rather than warnings for missing traits in [another PR](https://github.com/danielspaniel/ember-data-factory-guy/pull/264#issuecomment-266524766). I decided to spike this out. 

If this is merged it might be worth flagging this - maybe this change is bigger than a patch release. It's possible it will cause test failures for people where before they were seeing (but perhaps not noticing or acting on) warnings that were there all along.

### DIFF
--- a/tests/unit/mock-request-test.js
+++ b/tests/unit/mock-request-test.js
@@ -263,7 +263,7 @@ test("with errors in response", function(assert) {
     const done = assert.async();
 
     const response = { errors: { description: ['bad'] } };
-    const mock = mockFindRecord('profile', 1).fails({ response });
+    const mock = mockFindRecord('profile').fails({ response });
 
     FactoryGuy.store.findRecord('profile', 1)
       .catch((res)=> {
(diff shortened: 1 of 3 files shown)


tests/unit/mock-request-test.js
```diff
@@ -263,7 +263,7 @@ test("with errors in response", function(assert) {
     const done = assert.async();
 
     const response = { errors: { description: ['bad'] } };
-    const mock = mockFindRecord('profile', 1).fails({ response });
+    const mock = mockFindRecord('profile').fails({ response });
```
lorcan: Funnily this change exposed this 
> WARNING: [ember-data-factory-guy] You're trying to use a trait [1] for model profile but that trait can not be found.

danielspaniel: oooo .. that is so interesting .. that is vestige of past , the way it was used .. good catch ..

danielspaniel:
was there another thing you were going to use assert for .. you want to do that one and then I put that out in next 2.10 release ?

lorcan:
Yeah was hoping to complete https://github.com/danielspaniel/ember-data-factory-guy/pull/264 alongside but it got a little involved... It'll probably be next week before I get back to it.

PR #268, opened 2017-01-16
[WIP] use the new `store.isFragment` method

GavinJoyce:
We are currently relying on [`store.createFragment` raising an exception](https://github.com/danielspaniel/ember-data-factory-guy/blob/20ba57cafaa366b59e62fd9e1afef01d5d673c53/addon/model-definition.js#L41-L50) to detect if a model is a factory. This exception comes from an assertion which will be stripped in certain builds once https://github.com/ember-cli/rfcs/pull/50 is adopted.

I recently added a [`store.isFragment`](https://github.com/lytics/ember-data-model-fragments/pull/235) method to `ember-data-model-fragments`. This hasn't yet been released, but when it is it would be nice to start using it.

TODO:

 * [x] Update the `added in 'ember-data-model-fragments' v2.3.x` comment once it has been released

### DIFF
--- a/addon/model-definition.js
+++ b/addon/model-definition.js
@@ -39,6 +39,10 @@ class ModelDefinition {
    @returns {Boolean} true if it's a model fragment
    */
   isModelAFragment() {
+    if(FactoryGuy.store.isFragment) { //added in `ember-data-model-fragments` v2.11.0
+      return FactoryGuy.store.isFragment(this.modelName);
+    }
+
     try {
       if (FactoryGuy.store.createFragment) {
         return !!FactoryGuy.store.createFragment(this.modelName);


danielspaniel:
This is pretty great Gavin.
 Should have thought to do this myself because the other day I was looking at that method and was not loving it because the try catch was swallowing an exception ( mistake I was making ) in my test ( which made it hard / nearly impossible to detect ).
 I did not think to add what you did ( a nice method in the fragment addon ) .. so 'heck yeah' to this. 
Let me know as soon as that new version is ready.

GavinJoyce:
@danielspaniel ember-data-model-fragments 2.11.0 has been [released](https://github.com/lytics/ember-data-model-fragments/releases/tag/v2.11.0) so I think this is ready to go now

danielspaniel:
How should I handle this for people who have not updated to 2.11 fragments.
 I saw the thread in the fragments PR and yes, I should have used Fragment.detect, I just did not know about that method. 
 What I am thinking is that I could just copy what you are doing is that method

```javascript
  let type = store.modelFor(modelName);
  Fragment.detect(type); 
```

rather than use the method store.isFragment since then people are forced to update to 2.11 fragments .. know what I mean.  Thoughts?

GavinJoyce:
Thanks sounds fine @danielspaniel, feel free to close this PR if you go down that route

danielspaniel:
I will close this then @GavinJoyce , because I converted that method to use your code from that method and it works fine now. Much better .. so thanks again.

danielspaniel:
Release:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.10.1
 has the better implemented fragment checker ( copied from you )

PR #271, opened 2017-01-30
Update package.json to be the correct release number

patocallaghan:
I noticed the package.json doesn't mirror the release number


danielspaniel:
hmmm .. that is interesting .. wonder how that happened?  but good catch 😄

PR #281, opened 2017-04-06
jquery-mockjax from npm

### LINKED ISSUE #280: Move jquery-mockjax from bower dependency to npm dependency?
Hi, @danielspaniel!

Have you already considered moving `jquery-mockjax` from being a bower dependency to an npm dependency? `jquery-mockjax` is the last vestige of bower that I have left in my project. Considering the Ember community seems to be moving away from bower, I was curious if you've thought about it already.

Is it a matter of getting the time to work on it? Or are there some blockers I'm unaware of? I might be able to take a crack at it if the answer is the former instead of the latter. Thanks!

mblayman:
This branch converts `ember-data-factory-guy` to find `jquery-mockjax` from npm instead of bower. It does two primary things:
1. Update the blueprint to put `jquery-mockjax` into `package.json`.
2. Add `jquery-mockjax` to the vendor tree since node packages don't have the same status that bower components have.

This might be a breaking change as is since existing users won't have `jquery-mockjax` in their `node_modules` directory. One thought would be to forget adding `jquery-mockjax` to the blueprint and making it a dependency for the entire project.

I'm happy to update whatever needs updating as this is my first Ember addon contribution.

As far as testing goes, I did a manual test against my own project using a Git URL followed by `ember g ember-data-factory-guy`. Then I ran my tests and confirmed they still worked.

Fixes #280


mblayman:
Looks like the tests fail because they don't run `ember g ember-data-factory-guy` before starting. This feels like a good reason to make `jquery-mockjax` an npm dependency of the addon.

To me, it seems like ember-data-factory-guy could move the `jquery-mockjax` dependency from bower to npm and clean out the blueprint. I think that would get the addon itself one step closer to removing bower as well. 😄

danielspaniel:
Go for it. If you can get it working ( tests and all ) that would pretty awesome.

mblayman:
Yay! Tests are green. This is probably ready for review now.

danielspaniel:
only change was I put mockjax in dependencies and not devDependencies ( since it is a hard dependency ) .. but thanks for the effort .. :)

mblayman:
Awesome! Thanks a ton. I look forward to being bower-free. 😄

PR #287, opened 2017-05-10
Adding failing test to highlight meta information getting dropped from json requests

jeobrien:
refers to https://github.com/danielspaniel/ember-data-factory-guy/issues/267

I also am experiencing this issue - here is a failing test to demonstrate (the same test as the one I found the issue from). If you put in a debugger in response returned in the users route from the query you will see the meta keyword is undefined.

### DIFF
--- a/tests/acceptance/users-view-test.js
+++ b/tests/acceptance/users-view-test.js
@@ -1,5 +1,5 @@
 import {test} from 'qunit';
-import {buildList, makeList, mockFindAll} from 'ember-data-factory-guy';
+import {buildList, makeList, mockFindAll, mockQuery}  from 'ember-data-factory-guy';
 import moduleForAcceptance from '../helpers/module-for-acceptance';
 
 moduleForAcceptance('Acceptance | Users View');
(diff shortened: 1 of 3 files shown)


danielspaniel:
Nice one, thanks for sending the test for me to fiddle with .. will whip up a solution for you for the weekend.

jeobrien:
thanks @danielspaniel :)

danielspaniel:
You made a mistake @jeobrien in your test. ( the template was not correctly done )  

the template should have read 
```hbs
 {{#if nextPage}}
   <span class="meta">Next Page</span>
 {{/if}}
 {{#if previousPage}}
   <span class="meta">Previous Page</span>
 {{/if}}
```
 ( do you see the mistake you made? )
 
so actually the test works fine. And makes me see that the adding meta works perfectly ( as advertised ) .. therefore .. thanks for the extra test .. and I am going to close this PR and the issue until someone can find actual fail and I will reopen 

 But thank you very much for trying to highlight and explore the issue. This is a very fun feature and not one I want to see broken.

danielspaniel:
also .. another tip .. you could have written the route like this 

```javascript
  model: function() {
    return this.store.query('user', { page: 1 })
  },

  setupController(controller, users) {
    controller.set('previousPage', users.meta.previous);
    controller.set('nextPage', users.meta.next);
  }
```

danielspaniel:
And .. just for fun ( because I liked your test ) I moved your code to  test in users-search-test because mockQuery was used there .. so thanks again

jeobrien:
@danielspaniel Woops my mistake! Thanks for fixing that up, glad it was someone helpful at least :)

PR #295, opened 2017-06-06
correct widget order in Tip 1

mminkoff:
It's easy enough to see that it's a simple mistake, but better to avoid any confusion.


danielspaniel:
it never ceases to amaze me how many little mistakes like this one exists. Seems like every time I read the docs I see another one!  Thanks @mminkoff  for taking the time to lift your fingers up and down to fix this 👍

mminkoff:
It's a pleasure to do such a simple thing to support a project that does so much for me!

PR #299, opened 2017-08-02
Update ember-cli-babel

andrei-zhidkov:
Hello,

The current version of ember-cli-babel that is used in the project has quite restrictive version range. This causes some inconvenience when using Node 8, especially with yarn — ember-cli-babel@6.0.0 doesn't specify Node 8 as a supported engine, so every yarn command has to be accompanied with `--ignore-engines` flag. This also makes ember-cli commands like `ember init` impossible to complete, since there is no way to pass the required flag to yarn through ember-cli.

This pull request updates the ember-cli-babel to 6.7.0, the latest version available on npm at the moment, as well as sets a more permissive version range to avoid similar issues in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "dependencies": {
     "broccoli-funnel": "^1.0.1",
     "broccoli-merge-trees": "^1.1.1",
-    "ember-cli-babel": "6.0.0",
+    "ember-cli-babel": "^6.7.0",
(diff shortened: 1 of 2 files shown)


danielspaniel:
good idea, thanks for the PR 😄

danielspaniel:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v2.13.6   has this change 👍

PR #301, opened 2017-08-09
add failing test for factory inheritance

miguelcobain:
I think I've found a situation where the inheritance doesn't work as expected.
Basically when we use more than one level, it doesn't seem to work. In the test, the overridden property is undefined. It also seems to happen with properties that are defined in the parent factories (they end up being undefined as well).


danielspaniel:
That was a good find .. surprised I did not see that a while ago.

danielspaniel:
v12.13.7 has this fixed

PR #302, opened 2017-08-18
add travis testing for ember-data 2.14 breaking changes

mellatone:


danielspaniel:
now that ember data fixed polymorphic relationships, the factory guy tests now work for ED 2.14+

PR #305, opened 2017-09-18
Added controller unit test example in readme

jeffrey008:
When I was using  factory guy in controller unit test I didn't put things in `needs` array. Then I get `No model for 'post'` error. Hope this update can make less people get trapped.


danielspaniel:
Going to close this because new style of ember-qunit testing does not require needs anymore for any kind of test .. whooray!

PR #314, opened 2017-11-23
Add test and serializer to show failing tests with ember-data-model-fragments

jniechcial:
Addresses https://github.com/danielspaniel/ember-data-factory-guy/issues/313

Shows that after adding a serializer for a data-model-fragment it is being serialized properly in `build` step but is not being deserialized in `make` step of parent model.


danielspaniel:
In this case you would have to avoid using the returns({model}) and use returns({json}) like this:

```javascript 
test("Show employee by make(ing) a model ( with belongsTo fragment added manually ) and using returns with that model", async function(assert) {
  let firstName = 'Joe',
      lastName  = 'Black',
      name      = build('name', {firstName, lastName}).get(),
      employee  = build('employee', {name});

  mockFindRecord('employee').returns({json: employee});
  await visit('/employee/' + employee.get('id'));

  assert.ok(find('.name').text().match(`${firstName} ${lastName}`));
});
```

This is a really tricky problem, and what I appear to be doing in build is running the data through the serializer to make it look like a payload going to the server.  I seem to be assuming that the server payload going TO the server would be the same as the server payload going BACK to the server. 
This might be wrong assumption, but I am not sure how to fix it at the moment. 
For now, you can use the style I describe above where you send pure build payload in a mockFindRecord

jniechcial:
@danielspaniel True, I added four tests that show inconsistency. I know that for now it's not easy fix - so I think that it might be good idea to put more pressure in documentation about recommended approach.

IMO the docs should recommend using `build` everytime explicitly with `model-fragments` and json returning. What do you think? I think that at least couple of people could avoid confusion like me :) I can do it and create another PR if you agree.

danielspaniel:
That is good idea @jniechcial .. give me a day to try and figure this out though. It kind of bothers me so I am going to see what I can fix up.

danielspaniel:
I fixed this issue, so closing the PR

PR #315, opened 2017-11-27
Import pluralize from ember-inflector

krasnoukhov:
This is to address a deprecation warning:

```
DEPRECATION: Ember.String.pluralize() is deprecated. Please explicitly: import { pluralize } from 'ember-inflector'; [deprecation id: ember-inflector.globals]
```


krasnoukhov:
Closing in favor of #316, sorry for that

PR #319, opened 2017-12-01
Propose changes to documentation

jniechcial:
I suggest a couple of small changes to documentation to make it easier to understand that mixing `build` and `make` can have unpredictable results. I also underline that serializers are taken into consideration in `build` methods (what might be obvious, but I feel that it is easy to forget about).

Also small clean up in acceptance tests regarding model fragments.


danielspaniel:
thanks very much @jniechcial for docs and clean up 👍

PR #321, opened 2017-12-01
Fixes Issue 320 => mockGetRequest with model changes object type

gleseur:
The test is quite ugly but demonstrates the problem (see first commit)
Second commit fixes the issue and does not break any test, but it a bit surprising, so maybe this is necessary for some cases that are not documented in tests.

The test would need to be cleaned up before merge if ok on principe


danielspaniel:
thanks for the test ( it is not that bad actually .. i like it ) .. and will see about fixing it. 
and your fix is very interesting and I am surprised also. it probably is an incomplete fix, but that was neat-o  idea.

danielspaniel:
You were right. That was the perfect fix. Thanks for taking the time to fix it.  This is fixed now in  => v2.13.19

PR #322, opened 2018-01-15
Fixing broken links in the README

Oreoz:
Simply fixing a couple of dead links in the docs!


danielspaniel:
thanks @Oreoz .. I know this is small change, but the docs are important and I can't believe I did not see that before! 👍

PR #329, opened 2018-02-19
mergeDeep: Add `hasOwnProperty()` check

Turbo87:
Without this change our tests are running into errors like described in https://github.com/emberjs/ember.js/issues/16049, because computed properties are included in the for-in loop and the nested `mergeDeep()` call will try to clone `isDescriptor` and other internal properties of the computed property.

Related to https://github.com/danielspaniel/ember-data-factory-guy/issues/328

/cc @danielspaniel


danielspaniel:
Thanks for the fix => @Turbo87  🏅

PR #333, opened 2018-03-07
Non-polymorphic embedded hasMany works with AMS

pangratz:
It seems that at the moment there is only test coverage for embedded, polymorphic hasMany relationships - but not for non-polymorphic ones.

Non-polymorphic embedded hasMany works for rest and json serializer, but it doesn't with the ActiveModelSerializer.

The fix basically includes 3 changes in the codebase, which hopefully are comprehensible. I will add some comments for further explanation.

Happy to update this PR, split it up so it only contains the fix (and add test coverage in a separate PR). Also, let me know if there is something I've missed, but testing wise the fixes seem not to influence other parts, even though the 3 changes look more substantial...

### DIFF
--- a/addon/converter/fixture-converter.js
+++ b/addon/converter/fixture-converter.js
@@ -94,7 +94,7 @@ export default class {
   }
 
   transformRelationshipKey(relationship) {
-    let transformFn = this.getTransformKeyFunction(relationship.type, 'Relationship');
+    let transformFn = this.getTransformKeyFunction(relationship.parentType.modelName, 'Relationship');
     return transformFn(relationship.key, relationship.kind);
   }
 
@@ -135,7 +135,7 @@ export default class {
           attrName = attrOptions;
         }
       }
-      return attrName || keyFn.apply(this, [attribute, method]);
+      return attrName || keyFn.apply(serializer, [attribute, method, 'serialize']);
     });
   }
 
@@ -240,7 +240,7 @@ export default class {
   extractHasMany(fixture, relationship, parentModelName, relationships) {
     let hasManyRecords  = fixture[relationship.key],
         relationshipKey = this.transformRelationshipKey(relationship),
-        isEmbedded      = this.isEmbeddedRelationship(parentModelName, relationshipKey),
+        isEmbedded      = this.isEmbeddedRelationship(parentModelName, relationship.key),
         isLinks         = hasManyRecords && hasManyRecords.links;
 
     if (isLinks) {
--- a/tests/dummy/app/models/comic-book.js
+++ b/tests/dummy/app/models/comic-book.js
@@ -6,4 +6,5 @@ export default Model.extend({
   name: attr('string'),
   company: belongsTo('company'),
   characters: hasMany('person', { polymorphic: true }),
+  includedVillains: hasMany('villain'),
 });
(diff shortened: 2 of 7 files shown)


addon/converter/fixture-converter.js
```diff
@@ -94,7 +94,7 @@ export default class {
   }
 
   transformRelationshipKey(relationship) {
-    let transformFn = this.getTransformKeyFunction(relationship.type, 'Relationship');
+    let transformFn = this.getTransformKeyFunction(relationship.parentType.modelName, 'Relationship');
```
pangratz: Actually the serializer of the parent of the relationship should be asked for the key of the given relationship. I am actually quite surprised that this hasn't surfaced earlier, so maybe I am missing something?

addon/converter/fixture-converter.js
```diff
@@ -240,7 +240,7 @@ export default class {
   extractHasMany(fixture, relationship, parentModelName, relationships) {
     let hasManyRecords  = fixture[relationship.key],
         relationshipKey = this.transformRelationshipKey(relationship),
-        isEmbedded      = this.isEmbeddedRelationship(parentModelName, relationshipKey),
+        isEmbedded      = this.isEmbeddedRelationship(parentModelName, relationship.key),
```
pangratz: We need to pass the `key` of the relationship (instead of the transformed one), because otherwise the lookup on the `attrs` has will check for the wrong key. Since the `attrs` hash uses the relationship names as defined in the model, and not the keys which are in the payload...

---

Given

```js
attrs: {
  includedVillains: { embedded: 'always' }
}
```

we want to use `this.isEmbeddedRelationship(parentModelName, "includedVillains")` instead of `this.isEmbeddedRelationship(parentModelName, "included_villains")`.

addon/converter/fixture-converter.js
danielspaniel: Oddly I am using AMS and polymorphic relationships all over the place, so I would have thought so too.  Sometimes bugs just suddenly stand up and say 'notice me' .. or maybe you just got the perfect edge case here

addon/converter/fixture-converter.js
danielspaniel: tricky

danielspaniel:
neat o .. nice PR

addon/converter/fixture-converter.js
```diff
@@ -135,7 +135,7 @@ export default class {
           attrName = attrOptions;
         }
       }
-      return attrName || keyFn.apply(this, [attribute, method]);
+      return attrName || keyFn.apply(serializer, [attribute, method, 'serialize']);
```
pangratz: This change surprised me the most, as I would have thought that an incorrect `this` within a serializers' `keyForRelationship` / `keyForAttribute` would be something somebody would have stumbled over earlier 🤔 

Passing the 3rd `serialize` argument is needed so we ensure [this path](https://github.com/emberjs/data/blob/v3.0.2/addon/serializers/embedded-records-mixin.js#L135) is taken when the embedded records mixin is used.

pangratz:
Uff, that was quick. Thanks a bunch!! 🚀

addon/converter/fixture-converter.js
danielspaniel: I recall a year ago I was fumbling with this .. and I just don't remember what happened .. or I how this got to this point ( been too long ) so welcome the fresh eyes on this 👍

danielspaniel:
new patch version arriving soon

danielspaniel:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v3.0.1

PR #338, opened 2018-04-25
getPretender and setPretender functions as public exports

ryedeer:
Hello! This is the PR I promised to make some time ago. It allows to import the `getPretender` function to use the FactoryGuy's instance of Pretender for other tasks. Also it includes the `setPretender` function to let the FactoryGuy use the Pretender instance created by some other library.


danielspaniel:
Hi @ryedeer ,  remind me again why you/someone would want to setPretender ( can you think of a use case for that ? ) I am having trouble.  I guess i forgot our last conversation.

ryedeer:
The reason to have `setPretender`: only one instance of Pretenter is allowed to exist, and some third-party library that also uses Pretender can also instantiate it, so it would be nice to have a mechanism to use the instance that already exists instead of trying to create another one.

ryedeer:
Currently, we only use `getPretender`, so I cannot show a specific use case for `setPretender`. Well, I can just delete all traces of it from the pull request... It wouldn't be hard to implement it when it is really necessary.

danielspaniel:
I am not trying to be difficult and spoil a nice PR Oleg, I just want to make sure to only add things that actually would be useful, and not create things that no one would really use. So if you can remove that setPretender till you need it or someone asks for it that would make me happier.

ryedeer:
Yes, I understand it. Here's a commit that removes `setPretender`.

Unrelated to this PR: may I ask you to check my recent comment to the [issue 231](https://github.com/danielspaniel/ember-data-factory-guy/issues/231)? A similar problem exists again, and I cannot reopen that issue because it wasn't closed by me.

PR #339, opened 2018-04-26
Use run.join instead of run

Subtletree:
I was having an issue where doing lots of `makeList` in a test caused more than 10 runloops to be created which caused an infinite rendering invalidation error.

A similar issue was reported here
https://github.com/ember-redux/ember-redux/issues/64#issuecomment-267053666

And fixed here using `run.join` to only create new loops as needed
https://github.com/ember-redux/ember-redux/pull/69


danielspaniel:
Interesting issue .. never ran into it myself ( yet )

Subtletree:
Thanks!

Yeah likely something specific to my scenario but good as a safeguard either way.

PR #348, opened 2018-07-03
Fix issue with buildList and a count of 0

patocallaghan:
Fixes the following issue where passing `0` to `buildList` doesn't work correctly.

``` js
// before fix
buildList('article', 0, 'published', { folder: home, owner: 29 }).length => 2

// after fix
buildList('article', 0, 'published', { folder: home, owner: 29 }).length => 0
```

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -319,7 +319,7 @@ class FactoryGuy {
   buildRawList({name, number, opts, buildType = 'build'} = {}) {
     let definition = FactoryGuy.lookupDefinitionForFixtureName(name, true);
 
-    if (number) {
+    if (number != null) {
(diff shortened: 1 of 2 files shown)


lorcan:
There's similar logic in makeList: https://github.com/danielspaniel/ember-data-factory-guy/blob/1058950aedc66ef64c3e02d46b9f23bf364b103f/addon/factory-guy.js#L442-L444 

do we need to do anything there too?

patocallaghan:
Ah good catch @lorcan, I'll add a fix for that too 👍

patocallaghan:
@lorcan `makeList` had the same bug, thanks 👍

danielspaniel:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v3.1.5  has this fix

PR #349, opened 2018-07-03
Add postSetup hook

patocallaghan:
### Why?
I'd like some sort of hook which Factory Guy would run after set up has occurred for each test, i.e. `manualSetup`. Rather than have to add logic to each individual unit, integration and acceptance test and I'd like to be able to set it in a single place, for example in `tests/test-helper.js`. In my particular use case what I want to do is get access to the `Pretender` instance and set custom `unhandledRequest` or `erroredRequest` handlers.

### How?
I've added a new property `FactoryGuy#postSetup` which is set through the `settings` hash

``` js
FactoryGuy.settings({
  postSetup() {
    // my custom logic
  }
});

// in manualSetup
FactoryGuy.postSetup();
```

I'm not tied to the `postSetup` name so if you have any better suggestions? Also if you know of a way to achieve this without needing to add this PR I'd love to know.

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -24,12 +24,14 @@ class FactoryGuy {
    *
    * responseTime: 0 is fastest
    * logLevel: 0 is off, 1 is on
+   * postSetup: a hook to run after Factory Guy sets up
    *
    * @param logLevel [0/1]
    */
-  settings({logLevel = 0, responseTime = null} = {}) {
+  settings({logLevel = 0, responseTime = null, postSetup = null} = {}) {
     RequestManager.settings({responseTime});
     this.logLevel = logLevel;
+    this.postSetup = postSetup;
     return RequestManager.settings();
   }
 
--- a/addon/utils/manual-setup.js
+++ b/addon/utils/manual-setup.js
@@ -10,5 +10,8 @@ export default function(scope) {
   FactoryGuy.setStore(owner.lookup('service:store'));
   loadFactories();
   loadScenarios(owner);
+  if (FactoryGuy.postSetup) {
+    FactoryGuy.postSetup();
+  }
 }
 
--- a/tests/unit/factory-guy-test.js
+++ b/tests/unit/factory-guy-test.js
@@ -3,7 +3,7 @@ import { setupTest } from 'ember-qunit';
 import DS from 'ember-data';
 import Ember from 'ember';
 import FactoryGuy, {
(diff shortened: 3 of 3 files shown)


patocallaghan:
So I was hoping not to have to write our own wrapper function 😄 but I see how this could work. How would this interact with the initializer that calls `manualSetup`? Will there be any weirdness calling `manualSetup` twice in acceptance tests?

danielspaniel:
yes there would be weirdness Pat ( and is highly not recommended ) , but what is the use case for calling manual setup 2 times .. me confused

danielspaniel:
i can screenshare or something if you want to explain more or have odd use case ?

patocallaghan:
In acceptance tests isn't `manualSetup` called here https://github.com/danielspaniel/ember-data-factory-guy/blob/2bfac38cfc02afa5c876523157fff35d4638ff0f/app/initializers/ember-data-factory-guy.js#L13?

So if I created a separate helper that wrapped `manualSetup` and called that in my test, `manualSetup` would be called twice? Unless I'm misunderstanding how that initializer works?

danielspaniel:
holy crapoly Pat, I forgot that manualSetup was automatically called .. yikes.  I think that is especially for use with factory guy with scenarios. This should probably be changed so it is not called and you call manualSetup in acceptance tests yourself ( or setupFactoryGuy .. same idea ) .. let me ponder this for a moment and see if I can change this behaviour.

danielspaniel:
I keep forgetting to get to this @patocallaghan  .. sorry .. brain has been on vacation

patocallaghan:
@danielspaniel Closing this but would still love to hear your thoughts.

danielspaniel:
Hi Pat,  What I am thinking is that in acceptance tests manualSetup is not called  IF IF IF you use the new style tests , like this: 

```js
module('Acceptance | my-fun-page', function(hooks) {
  setupApplicationTest(hooks);
  setupFactoryGuy(hooks);
  ... etc 
```

so I am thinking that you are using old style tests .. right? 
  And if so .. you could update your tests to use the new styling? and avoid manualSetup being called automatically. 
  Then you could write a function that all tests use that has manualSetup in it PLUS the post setup you want and voila .. all set ..

 I think the reason I was hesitating is because the concept just seemed too klugy. Granted factory guy forced you to do that, but not anymore ( with new style tests ) 

 but let me know if I answered question properly. I don't want to just say .. oh upgrade your tests ( cause I know that is alot of work )

Issue #353, opened 2018-07-29
Cannot read property 'fetch' of undefined


herzzanu:
Upgrading to the latest version gets me this error. 

![image](https://user-images.githubusercontent.com/2959519/43370366-0deef59e-937e-11e8-8e14-5425fa26a8f3.png)

Both in an existing app with Ember 3.0 and in a brand new app with Ember 3.3.

herzzanu:
Probably duplicate of #352

danielspaniel:
v3.3.1 fixes this ( I think ) https://github.com/danielspaniel/ember-data-factory-guy/pull/354

danielspaniel:
hmmm .. not sure what to say.  Can you try and noodle around with package.json and see if you can fix it?

herzzanu:
Seems this one https://github.com/danielspaniel/ember-data-factory-guy/pull/388 fixes it.

danielspaniel:
super .. will merge it and make new version in a moment  .. thanks for PR

herzzanu:
Awesome, you rock 🎉

Issue #352, opened 2018-07-27
'WhatWGFetch' is not defined in pretender.js


Gaurav0:
The latest versions of pretender brought in by this addon, v2.1.1, has a dependency of @xg-wang/whatwg-fetch, which is not imported to the browser by this addon. As a result, during tests, one gets an error that WhatWGfetch is not defined.

The solution is to import all dependencies required by pretender by reading its package.json.

As a workaround, you can use the resolutions field in package.json to specify pretender v2.0.0.

danielspaniel:
You want to make a PR  for that @Gaurav0 ?

danielspaniel:
v3.3.1 has fix from this PR https://github.com/danielspaniel/ember-data-factory-guy/pull/354

PR #354, opened 2018-07-31
Fix pretender

Exelord:
This PR will fix pretender issue.
#353 #352


danielspaniel:
was the fix that easy?  gee .. me surprised .. but happy ? :) 
thanks @Exelord

Exelord:
Try to not use `^` in dependencies. People are making mistakes :/

PR #359, opened 2018-08-15
Update testem.js

runspired:
The above flags improve the stability of using Chrome, and will be required for ember-data to be able to run CI tests that include ember-data-factory-guy tests


danielspaniel:
me like :)

PR #362, opened 2018-08-25
Allow `make` to return an instance of `ModelFragment`

patocallaghan:
Currently if you use `FactoryGuy#make` to generate a `ModelFragment` it returns a json hash of the properties of the fragment. In keeping with the consistency of using `Factory#make` for a regular Ember Data record, I think it should return an actual model instance instead, in this case a `ModelFragment`.

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -379,7 +379,7 @@ class FactoryGuy {
         fixture     = this.buildRaw(assign(args, {buildType: 'make'}));
 
     if (this.isModelAFragment(modelName)) {
-      return fixture;
+      return this.store.createFragment(modelName, fixture);
     }
 
     let data  = this.fixtureBuilder(modelName).convertForMake(modelName, fixture),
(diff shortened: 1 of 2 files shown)


addon/factory-guy.js
```diff
@@ -379,7 +379,7 @@ class FactoryGuy {
         fixture     = this.buildRaw(assign(args, {buildType: 'make'}));
 
     if (this.isModelAFragment(modelName)) {
-      return fixture;
+      return this.store.createFragment(modelName, fixture);
```
patocallaghan: I'm assuming that I don't need to `this.store.normalize` the fixture as this has already been done inside `buildRaw`?

patocallaghan: `Store#createFragment` is the API from [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments)

patocallaghan:
@danielspaniel any idea when you'll be publishing a new release?

danielspaniel:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v3.3.5   is ready .. thanks for reminder

patocallaghan:
Awesome, thanks @danielspaniel! 👏

PR #363, opened 2018-08-28
Create fragment inside a run loop

patocallaghan:
After pulling the latest version with [my fragment fix](https://github.com/danielspaniel/ember-data-factory-guy/pull/362) into my app we started getting asynchronous errors when creating fragments directly using `make`. I've updated to create the fragment inside a run loop. This is similar to what we do when pushing records into the regular store just below these lines.

``` js
model = join(() => this.store.push(data));
```

### Before
image

### After
image

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -379,7 +379,7 @@ class FactoryGuy {
         fixture     = this.buildRaw(assign(args, {buildType: 'make'}));
 
     if (this.isModelAFragment(modelName)) {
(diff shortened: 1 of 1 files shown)


danielspaniel:
thanks for fixing that Pat

danielspaniel:
https://github.com/danielspaniel/ember-data-factory-guy/releases/tag/v3.3.6

patocallaghan:
Thanks! 🙌

PR #368, opened 2018-11-15
Add some additional tests for `mock`

patocallaghan:
@danielspaniel after our discussions on Discord I added some tests which verified some `mock` behaviour to me. No harm in adding them to the test suite.


danielspaniel:
thanks for confirming that @patocallaghan  :)

PR #372, opened 2019-01-23
add import statement for `setupFactoryGuy`

pangratz:
Otherwise it's not immediately obvious where this utility is imported from.


danielspaniel:
good point 👍

PR #375, opened 2019-02-13
mockDelete should also add model to fake snapshot

artinboghosian:
In regards to https://github.com/danielspaniel/ember-data-factory-guy/issues/374, we found a simpler workaround for the time being to have mockDelete support passing the model as record in the snapshot (since DS.Snapshot also passes this value), which we can utilize instead of `snapshot.attr` in our case. This was simpler than trying to mimic all functionality of DS.Snapshot (which is in -private namespace). 

This is what `mockUpdate` does as well and since `mockDelete` is generally dealing with a persisted record in the same way as `mockUpdate` they should behave the same in regards to generating a fake snapshot.

I simply copied over the `makeFakeSnapshot` method from `mock-update-request.js`. I don't know if there is a better way. I thought about a mixin like `AttributeMatcher` or `MaybeIdUrlMatch` but that would mean inheriting from multiple classes (which is not a pattern in the repo). The other options I thought of was adding a util method to house the logic, but this didn't feel like a util method.

### DIFF
--- a/addon/mocks/exposed-request-functions.js
+++ b/addon/mocks/exposed-request-functions.js
@@ -371,20 +371,19 @@ export function mockUpdate(...args) {
  @param {String} id optional id of record to delete
  */
 export function mockDelete(...args) {
-  let modelName, id;
+  let model, modelName, id;
 
   if (args[0] instanceof Model) {
-    let record = args[0];
-    modelName = record.constructor.modelName;
-    id = record.id;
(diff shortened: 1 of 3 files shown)


danielspaniel:
I think your idea was good. Thanks for the PR and the brainpower.

artinboghosian:
@danielspaniel thank you for quick turn around.

PR #380, opened 2019-03-04
Tolerate non-string request bodies

sduquej:
Fixes https://github.com/danielspaniel/ember-data-factory-guy/issues/379 by relaxing the assumption that `requestBody` is always a stringified JSON object. In the cases where it isn't, we can leave it untouched and pass it down to the matcher function so that it determines what to do.

### DIFF
--- a/addon/mocks/attribute-matcher.js
+++ b/addon/mocks/attribute-matcher.js
@@ -1,4 +1,5 @@
 import { assert } from '@ember/debug';
+import { typeOf } from '@ember/utils';
 import FactoryGuy from '../factory-guy';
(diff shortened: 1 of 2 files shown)


danielspaniel:
i think this was good idea

sduquej:
Thanks @danielspaniel! Would it be possible to release a new version that includes this?

danielspaniel:
yes, normally i would have done that right away, but i was working on a branch that i .. blah blah .. anyway .. yes, i get that for you asap ( by tomorrow AM  latest )

sduquej:
Thanks for the quick turnaround, appreciate it!

danielspaniel:
v3.6.6 has your fix

PR #389, opened 2019-06-10
Example of confusing error message

nolaneo:
image

```
Pretender intercepted GET /users?rightParams=false but encountered an error: Nothing returned by handler for /users?rightParams=false. Remember to `return [status, headers, body];` in your route handler
```


danielspaniel:
This is a very good point. I have had this happen to me and I know why it happens now, but I did not know at first and it took a while to figure out.

Techn1x:
https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/390#issuecomment-2875805368

PR #392, opened 2019-06-18
Matching arrays in query params - fixes #391

### LINKED ISSUE #391: Matching arrays in query params
Hello Daniel!

I've recently upgraded FactoryGuy from 3.4 and found that some of my tests are now failing; namely, the few tests that have arrays in query params for a mock. The mocks defined like `mockQuery('user').withSomeParams(by_roles: ["basic.vendor.1"])` are now failing with a message like `Error: Pretender intercepted GET http://example.com/api/v1/users?by_roles%5B%5D=basic.vendor.1 but encountered an error: Nothing returned by handler...` (which is pretty vague but I see there's another issue for it).

I checked the changes in the code and found the cause of the problem: `paramsMatch` method in `MockRequest` uses `toParams` for the defined mock params [but doesn't apply it to the actual query params](https://github.com/danielspaniel/ember-data-factory-guy/blob/master/addon/mocks/mock-request.js#L135-L140). To deal with the situation, I've made a quick patch that applied `toParams` to the both sides:

```
  paramsMatch(request) {
    if (!isEmptyObject(this.someQueryParams)) {
      return isPartOf(toParams(request.queryParams), toParams(this.someQueryParams));
    }
    if (!isEmptyObject(this.queryParams)) {
      return isEquivalent(toParams(request.queryParams), toParams(this.queryParams));
    }
    return true;
  }
```

It fixes the array problem and doesn't break any other tests, including the tests for the FactoryGuy itself. So I think it would be a good idea to use this update in the library.

Thank you!

ryedeer:


danielspaniel:
thanks again Oleg

PR #394, opened 2019-06-20
Fix relationship key transformation

misterbyrne:
This fixes https://github.com/danielspaniel/ember-data-factory-guy/issues/393
Not sure if I need to keep the `parentType` property there.

Could also do with adding another test for a serializer which includes a custom relationship key.

### DIFF
--- a/addon/converter/fixture-converter.js
+++ b/addon/converter/fixture-converter.js
@@ -99,8 +99,8 @@ export default class FixtureConverter {
   }
 
   transformRelationshipKey(relationship) {
-    let {parentType} = relationship,
-        type         = parentType && parentType.modelName || relationship.type,
+    let {parentModelName, parentType} = relationship,
+        type         = parentModelName || parentType && parentType.modelName || relationship.type,
         transformFn  = this.getTransformKeyFunction(type, 'Relationship');
     return transformFn(relationship.key, relationship.kind);
   }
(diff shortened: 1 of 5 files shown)


danielspaniel:
Yes, actually .. another test would be nice ?  to test the scenario you mentioned specifically.  think you can do it?

misterbyrne:
Cool. I've added another test. I tried to minimise the changes I had to make to the dummy models & serializers. Let me know if you think I should do something else instead. Thanks.

addon/converter/fixture-converter.js
```diff
@@ -99,8 +99,8 @@ export default class FixtureConverter {
   }
 
   transformRelationshipKey(relationship) {
-    let {parentType} = relationship,
-        type         = parentType && parentType.modelName || relationship.type,
+    let {parentModelName, parentType} = relationship,
```
misterbyrne: Is `parentType` still a property `relationship`? Possibly worth keeping this as a fallback for backwards compatibility if it was.

danielspaniel: your right, good to keep for backward compat,  ( not absolutely necessary because I make sure every version 3.8 , etc works with 3.8 ED ) but just a nice to have.

danielspaniel:
I will wait for other PR to make new patch version ( but thanks for this one Adam 😃

danielspaniel:
v 3.9.3 has this PR and the next one you made

misterbyrne:
Cool! Thanks for responding so quickly. Looking forward to trying this out in our main app as soon as I can.

PR #395, opened 2019-06-21
Fix .withParams for array parameters

misterbyrne:
A `MockAnyRequest` currently won't correctly match requests when the `.withParams` method is used with array parameters.

### DIFF
--- a/addon/mocks/mock-any-request.js
+++ b/addon/mocks/mock-any-request.js
@@ -1,6 +1,6 @@
 import MockRequest from './mock-request';
 import {
-  isEmptyObject, isEquivalent, isPartOf, paramsFromRequestBody, toGetParams, toParams
+  isEmptyObject, isEquivalent, isPartOf, paramsFromRequestBody, param, toParams
 } from "../utils/helper-functions";
 
 export default class MockAnyRequest extends MockRequest {
@@ -47,7 +47,7 @@ export default class MockAnyRequest extends MockRequest {
 
   _tryMatchParams(request, handlerParams, comparisonFunction) {
     if (this.type === 'GET') {
-      return comparisonFunction(toGetParams(request.queryParams), toGetParams(handlerParams));
+      return comparisonFunction(param(request.queryParams), param(handlerParams));
     }
 
     if (/POST|PUT|PATCH/.test(this.type)) {
--- a/addon/mocks/mock-any-request.js
+++ b/addon/mocks/mock-any-request.js
@@ -53,7 +53,7 @@ export default class MockAnyRequest extends MockRequest {
     if (/POST|PUT|PATCH/.test(this.type)) {
       const requestBody   = request.requestBody,
             requestParams = paramsFromRequestBody(requestBody);
-      return comparisonFunction(requestParams, toParams(handlerParams));
+      return comparisonFunction(toParams(requestParams), toParams(handlerParams));
     }
 
     return false;
--- a/addon/utils/helper-functions.js
+++ b/addon/utils/helper-functions.js
@@ -25,38 +25,6 @@ export function paramsFromRequestBody(body) {
   return body;
 }
 
-/**
- *
- * @param obj
- */
-export function toGetParams(obj) {
-  let convertedParams = Object.assign({}, obj);
-  Object.keys(convertedParams).forEach(key => {
-    let value = convertedParams[key];
-    if (typeOf(value) === 'array') {
-      value.forEach((v,i) => {
-        convertedParams[`${key}[${i}]`] = String(v)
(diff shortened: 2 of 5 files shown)


tests/helpers/utility-methods.js
```diff
@@ -9,14 +9,13 @@ import { ActiveModelSerializer } from 'active-model-adapter';
...
+  if (method === 'GET') {
+    url = [url, param(params)].join('?');
   } else {
-    const body = JSON.stringify(params)
-    return fetch(url, {body, method}).then(r => {
-      return r._bodyText ? r.json() : null;
-    });
+    body = JSON.stringify(params);
   }
+  return fetch(url, {body, method}).then(r => r._bodyText ? r.json() : null);
```
misterbyrne: I've moved this around so that it handles requests without a body, regardless of the method used.

tests/unit/mocks/mock-request-test.js
```diff
@@ -13,6 +13,18 @@ module('MockRequest', function(hooks) {
   setupTest(hooks);
   inlineSetup(hooks, serializerType);
 
+  module('#withParams', function() {
+    test('works with arrays', async function(assert) {
```
misterbyrne: This already passes. I guess this is because `MockRequest` is mostly used only for `GET` requests. I'm guessing that it might break for `POST` requests (which would be unusual but could happen with custom adapters). It might be difficult to setup this type of scenario and I think there's still value in this PR without doing this.

danielspaniel: Adam, you could write a test for the mock any with POST request ( taking array params )  because I use mock any with post / put alot myself so that would be useful ( if they don't already exist of course .. which i don't think they do )

tests/helpers/utility-methods.js
danielspaniel: good idea

tests/unit/mocks/mock-request-test.js
misterbyrne: Hi. I did add a test for mock any with `GET` / `POST` / `PATCH` and `PUT` [here](https://github.com/danielspaniel/ember-data-factory-guy/pull/395/files#diff-2da6df751e64bdd490e7704016192aa6R223).

The test this comment is on (`mock-request-test.js`) is a test for `mock-store-request`. I don't think there's a good way right now to mock a `POST` request `withParams` for this type of request.

misterbyrne: In fact, I think I'll delete this test altogether because it's not relevant.

danielspaniel: yes, your last round em up and test em all iterator GET, PUT, POST , PATCH  was good idea.
  Are you trying to test every type of param (  string , array , hash ) or just array and string.  If so is it worth doing hash type ( or did I already do that elsewhere? )

tests/unit/mocks/mock-any-test.js
```diff
@@ -200,4 +200,27 @@ module('MockAny', function(hooks) {
     assert.equal(fooMock.timesCalled, 1, 'fooMock#timesCalled is called once');
     assert.equal(barMock.timesCalled, 1, 'barMock#timesCalled is called once');
   });
+
+  function testWithParamsForMethod(method) {
+    test(`#withParams works with complex parameters for ${method} requests`, async function(assert) {
+      const url     = '/api/post/some-stuff',
+            dataFoo = {foo: 'foo', array: ['a', { b: 123 }, 'c'], obj: { a: 321 } },
```
misterbyrne: I've changed this test to include objects, arrays, and arrays which include objects.

addon/mocks/mock-any-request.js
```diff
@@ -47,13 +47,13 @@ export default class MockAnyRequest extends MockRequest {
 
   _tryMatchParams(request, handlerParams, comparisonFunction) {
     if (this.type === 'GET') {
-      return comparisonFunction(toGetParams(request.queryParams), toGetParams(handlerParams));
+      return comparisonFunction(param(request.queryParams), param(handlerParams));
```
misterbyrne: I had to make this change to get the test with the object params to pass.

addon/utils/helper-functions.js
```diff
@@ -25,38 +25,6 @@ export function paramsFromRequestBody(body) {
   return body;
 }
 
-/**
- *
- * @param obj
- */
-export function toGetParams(obj) {
```
misterbyrne: This wasn't used any more after my changes

danielspaniel:
awesome Adam, thanks for being diligent and not just settling for any slop that works.

PR #410, opened 2019-11-18
Fix snapshot record for Get requests

cinkonaap:
Related issue: https://github.com/danielspaniel/ember-data-factory-guy/issues/409
Problem:

While developing one of the tests in my app it turned out that when creating mocked find operation from already mocked model, like:

```
const user = make('user');
mockFindRecord(user);
```

the Snapshot object never had the `record` set to this model.


danielspaniel:
nice work @cinkonaap  and @springerigor

kbiedrzycki:
@danielspaniel any chance to publish new version including these changes soon?

danielspaniel:
v3.9.5 is now released with this fix

PR #411, opened 2019-12-16
Highlight ability to use pretender's passthrough feature

Subtletree:
I searched the repo for `data url` trying to find how to ignore them. The search turned up nothing so adding this to the readme could save someone time in the future.

Awesome that this is possible : )


danielspaniel:
thanks for helping with docs Ryan

PR #416, opened 2020-05-07
fix error when adapter modifies queryParams

BryanCrotaz:
Clones the query params before handing off to the adapter in case the adapter modifies it


danielspaniel:
new version ->  v3.9.6 has this patch in it.   Nice work by the way

Issue #427, opened 2020-09-08
Bump pretender to 3.4.3


jcardali:
Would like to bump pretender in a project that also uses ember-data-factory-guy, which currently causes a version mismatch because the highest that ember-data-factory-guy's pretender can go up to is 3.0.5, which is [nine months old](https://www.npmjs.com/package/pretender)

According to [pretender's changelog](https://github.com/pretenderjs/pretender/blob/master/CHANGELOG.md), there have been no breaking changes since 3.0.1.

~~I was able to clone the repo and do the bump locally, and all tests pass, however I'm unable to push the branch for some reason.~~

patocallaghan:
Released in 3.9.8

Issue #387, opened 2019-06-04
Cannot read property 'fetch' of undefined


herzzanu:
Seems there's again an issue with pretender.

![image](https://user-images.githubusercontent.com/2959519/58918996-d007e580-872c-11e9-83b2-44681f248555.png)

We had this type of issue before https://github.com/danielspaniel/ember-data-factory-guy/issues/353 and it was solved. Seems the pretender.js in package.json uses again the `^` and then it happens that we stumble upon these kind of issues without any warning 😕

danielspaniel:
v3.9.1 is ready :)

herzzanu:
Awesome 👏 Thanks!

PR #429, opened 2020-09-17
Bump pretender to 3.4.3

jcardali:
See https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/427

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "broccoli-merge-trees": "^3.0.2",
     "ember-auto-import": "^1.3.0",
     "ember-cli-babel": "^7.4.0",
-    "pretender": "~3.0.1",
+    "pretender": "^3.4.3",
     "whatwg-fetch": "^3.0.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -4373,9 +4373,10 @@ extglob@^2.0.4:
     snapdragon "^0.8.1"
     to-regex "^3.0.1"
 
-fake-xml-http-request@^2.0.0:
-  version "2.0.0"
-  resolved "https://registry.yarnpkg.com/fake-xml-http-request/-/fake-xml-http-request-2.0.0.tgz#41a92f0ca539477700cb1dafd2df251d55dac8ff"
+fake-xml-http-request@^2.1.1:
+  version "2.1.1"
+  resolved "https://registry.yarnpkg.com/fake-xml-http-request/-/fake-xml-http-request-2.1.1.tgz#279fdac235840d7a4dff77d98ec44bce9fc690a6"
+  integrity sha512-Kn2WYYS6cDBS5jq/voOfSGCA0TafOYAUPbEp8mUVpD/DVV5bQIDjlq+MLLvNUokkbTpjBVlLDaM5PnX+PwZMlw==
 
 fast-deep-equal@^1.0.0:
   version "1.1.0"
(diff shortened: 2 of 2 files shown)


jcardali:
@patocallaghan @danielspaniel

patocallaghan:
Thanks @jcardali. Although I'm not sure if it's so straightforward. It appears @herzzanu had issues before when bumping the minor version of pretender.js in https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/387. @herzzanu is there an easy way to replicate this error so we can figure out if upgrading pretender.js works?

package.json
```diff
@@ -26,7 +26,7 @@
     "broccoli-merge-trees": "^3.0.2",
     "ember-auto-import": "^1.3.0",
     "ember-cli-babel": "^7.4.0",
-    "pretender": "~3.0.1",
+    "pretender": "^3.4.3",
```
herzzanu, line 29: Why not `~3.4.3` ?

herzzanu:
> Thanks @jcardali. Although I'm not sure if it's so straightforward. It appears @herzzanu had issues before when bumping the minor version of pretender.js in #387. @herzzanu is there an easy way to replicate this error so we can figure out if upgrading pretender.js works?

@jcardali @patocallaghan I guess testing this updated version in an ember app should tell the truth if it's ok or not.

package.json
patocallaghan: Because using `^` says we should be able to use all Pretender versions up until 4.0.0 rather than `~` which will only allow up until `3.5.0`. As Pretender uses semver there shouldn't be any breaking changes until 4.0. Also all the other dependencies use `^`.

patocallaghan:
Thanks @herzzanu, I'll try it out in a new ember app. I just thought you'd come across something specific to make the bug occur 👍

package.json
herzzanu: I know but remember this is how we solved https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/387. What I suggested would help keep it "under control". We can intervene and bump the version again if needed anyway. WDYT?

herzzanu: Also same fix was made here previously https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/354

patocallaghan:
@jcardali any chance you could regenerate the `yarn.lock` file with this dependency update please?

package.json
patocallaghan: @herzzanu I'd prefer to leave it with `^` as I think it's better to have looser dependencies for host apps. Mirage.js also uses pretender with `^`. I know there's been issues in the past with stability but I do seem to remember there was a large chunk of refactoring happening at that time (they were introducing fetch into Pretender) and it appears things are a lot more stable now.

yarn.lock
```diff
@@ -4373,9 +4373,10 @@ extglob@^2.0.4:
     snapdragon "^0.8.1"
     to-regex "^3.0.1"
 
-fake-xml-http-request@^2.0.0:
-  version "2.0.0"
-  resolved "https://registry.yarnpkg.com/fake-xml-http-request/-/fake-xml-http-request-2.0.0.tgz#41a92f0ca539477700cb1dafd2df251d55dac8ff"
+fake-xml-http-request@^2.1.1:
+  version "2.1.1"
+  resolved "https://artifactory.bettermg.com/artifactory/api/npm/npm/fake-xml-http-request/-/fake-xml-http-request-2.1.1.tgz#279fdac235840d7a4dff77d98ec44bce9fc690a6"
```
patocallaghan, line 4378: @jcardali it looks like your using an internal registry here which is causing CI to fail

jcardali: @patocallaghan Sorry about that, fixed now

patocallaghan:
Thanks @jcardali 🙌 Released in [v3.9.8](https://github.com/adopted-ember-addons/ember-data-factory-guy/releases/tag/v3.9.8)

package.json
herzzanu: Sounds good to me 👍

PR #433, opened 2021-02-18
Migrate CI to use Github Actions

patocallaghan:
Part of https://github.com/adopted-ember-addons/program-guidelines/issues/18

Migrate from Travis to Github Actions. This is an extremely simplified migration of CI to Github Action. I've tested it works fine on the branch

image

It turned out the original Travis config didn't use Ember Try properly as it always ran the tests using `ember test`. I went down a bit of a rabbit hole of trying to use ember-try properly but got issues due to the repo's usage of Node 8. I'm going to be upgrading the addon soon-ish so I'll upgrade the CI setup as part of that.

/cc @jelhan


jelhan:
It seems as if you needed to write the workflow manually. Would be interested to hear why. Maybe I can extend create-github-actions-setup-for-ember-addon to fit better for similar use cases.

Not sure if you were aware but you can customize the configuration used with a small hack:

1. Let the script create the workflow as usually.
2. Edit the configuration persisted at top of the workflow. E.g. delete ember try scenarios.
3. Run the script again. It will pick up configuration of the existing workflow and update it accordingly.

I plan to ease this by adding an interactive mode to the script. It would allow a user to customize the configuration before it is applied.

patocallaghan:
So it turns out the the project didn't have ember-try as a dependency. It must have been removed at some stage but the configuration stayed around. When I tried to run the "yarn create" command it blew up because it couldn't find ember-try. At that stage I just opted to take a copy of the model-fragments one and chop it down to size. 

I'm aiming to upgrade the addon soon so I'll revisit the ember try situation then.

jelhan:
I see. In that case it's very unlikely that create-github-actions-setup-for-ember-addon works even with manual adjustments.

But you can run it later after upgrading and fixing ember-try. It should work than out of the box.

Issue #417, opened 2020-05-18
Compatibility with Ember Data Model Fragments new refactor


patocallaghan:
Ember Data Model Fragments hasn't worked with versions of Ember Data since 3.13+. This is being tracked on their repo in https://github.com/lytics/ember-data-model-fragments/issues/338

To unblock Ember Data upgrades the internals of Model Fragments are being completely refactored to use Ember Data's `RecordData` API. The ongoing work is happening in https://github.com/lytics/ember-data-model-fragments/pull/360/ and they are close to cutting an alpha release.

I tried out the branch in our codebase but we got a load of exceptions from within Model Fragments generated by Factory Guy. It would be great to make sure that Factory Guy is still compatible with this refactored Model Fragments. [I'm going to link a Draft PR]

danielspaniel:
Nice to hear from you Pat. It's been a while. 
I am sorry to say that I will have to leave that update work to someone else ( maybe you folks at intercom can do it ) because I am not using ember or factory guy anymore. I started using clojure ecosystem a few years ago and never looked back.  
 Hopefully you can round up the folks to make this happen.

patocallaghan:
@danielspaniel likewise 😃Its been too long! Yep, I'm happy to dig into it and try fix the issues.

Thinking longer term, considering you're no longer looking to maintain this addon, would you consider putting it for adoption to the [Ember Adopted Addons](https://github.com/adopted-ember-addons) organisation? That's not so pressing an issue I'd be happy to open another issue to discuss that if you think it's worth it.

danielspaniel:
OOoo .. that is good idea Pat, I will look into that adoption thing.

patocallaghan:
I looked into this and I've been able to reproduce an issue where the latest ember-data-factory-guy doesn't work with Ember Data 3.16 and Ember Data Model Fragments 5.0. I'm not entirely sure if the problem is specific to Factory Guy as I've also been able to replicate in our app just clicking but not in this specific repo yet.

I have a test reproduction in this repo here https://github.com/patocallaghan/model-fragments-bug. The specific commit with the changes is in https://github.com/patocallaghan/model-fragments-bug/commit/10dfd71c2244e39a725964e3f85c38af66eebae8

The following test `Unit | Model | grandparent` throws the following error:

```
Error: Assertion Failed: You can only add 'child' fragments or object literals to this property
    at Object.assert (http://localhost:58354/assets/vendor.js:36148:15)
[...]

PR #449, opened 2021-10-04
[BREAKING] Fixes incompatibility with Ember Data Model Fragments 5.0.0-beta+ and Ember Data 3.13+

patocallaghan:
Fix for #417

### Background
While Factory Guy's primary use case is to generate Ember Data models it also supports the ability to generate [Ember Data Model Fragments](https://github.com/adopted-ember-addons/ember-data-model-fragments). Since Ember Data 3.13+ though there have been a number of compatibility issues between Model Fragments & Ember Data which have been outlined in detail in https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/338. Previously trying to bump Ember Data in this repo resulted in all tests containing fragments to fail. At this stage the Model Fragments compatibility issues have been largely resolved with a big internal refactor to use `RecordData` instead (see [changelog](https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/master/CHANGELOG.md#v500-beta0-may-28-2020)) so I think it's time we try and upgrade the Ember Data again.

### The problem
One issue we're still seeing though though is how Factory Guy generates Fragments using `make`. The `make` helper allows you to push records directly into the store rather than need to use the `push`, `pushPayload` or `createRecord` APIs to bootstrap your models. A common pattern when using `make` is to compose your test models with other models using `make`. For example take the following `person` Ember Data model which contains a fragment:

```js
// models/user.js
export default class User extends Model {
  @fragment('info') info
}

// models/info.js
export default class Info extends Fragment {
  @attr name
}
```

This can be created in a test by doing the following (once you've created your factories correctly). Here we create a user model and pass it an `info` fragment model.

```js
let user = make('user', {
  info: make('info'),
});
```

In Ember Data 3.16+ it appears passing a fragment to a model like above causes the properties of `info` to be `undefined` rather than the values which were originally on the model.

```js
console.log(info.name); //=> "Jon Snow" 
console.log(user.info.name); //=> undefined
```

In `FixtureConverter` we use `store.push` to create the models. If you follow the calls down the internals it appears `store.push` accepted any fragment instance as-is and just set it as a property on the model. At some point in Ember Data 3.16+ `store.push` was hardened to _only_ accept JSON (which in fairness is what its public API states). I think it was a "happy accident" the fact it actually accepted a generated model fragment before and "just worked".

This was pretty much confirmed by @runspired on Discord
image

### Solution
When generating fixtures which are passed fragments (or fragment arrays) as properties we must pass their serialized form to `store.push` rather than passing them as instances.

### What is the breaking change?
The reason I"m calling this out as a breaking change is in some circumstances this will cause tests to fail if you have test assertions being run on the external fragment passed to `make`. The external fragment and the one generated on the model are no longer the same object.

```js
let info = make('info', { name: 'Jon' } );
let user = make('user', { info });

// < Ember Data 3.16 this was the same object. Here the actual fragment was accepted by `store.push` as a valid property
console.log(info == user.info); // true

// Ember Data 3.16+ are no longer the same object. The passed-in fragment is used to generate the raw JSON data passed to `store.push`.
console.log(info == user.info); // false
```


tests/helpers/utility-methods.js
```diff
@@ -129,7 +130,13 @@ export function containerSetup(application, serializerType) {
     let store = application.lookup('service:store');
 
     let adapterType = serializerType === '-json' ? '-rest' : serializerType;
-    let adapter = application.lookup('adapter:' + adapterType);
+    let adapter;
```
patocallaghan, line 133: These are test-only changes but it appears that doing the following out-of-the-box works. I need to register the RESTAdapter similar to the other adapters a few lines above.

```
application.lookup('adapter:-rest');
```

22a (approved):

patocallaghan:
As an extra data point to the stability of these changes I've ran these changes against the Intercom test suite, which has thousands of tests and makes heavy use of Factory Guy and Model Fragments, and CI is green 🟢.

PR #452, opened 2021-11-03
allow match() to be used for any mock request

dwickern:
Currently only `mockCreate` and `mockUpdate` support `match()`. This PR adds `match()` for the other kinds of mock requests. Request bodies can get complex so it's nice to have a callback where you can do assertions etc.

For the implementation I basically merged `AttributeMatcher` into `MockRequest`.


patocallaghan:
Thanks for working on this @dwickern. I've just kicked off CI to run the tests.

Any chance we could add some updates to the existing README which outlines how you'd use this?

Techn1x:
Thanks a bunch for this. I find this a worthwhile change for two reasons;
* i've found myself wanting of `match()` on other mock calls as well
* this code is actually simpler and gets the code away from the "mixin-like" behaviour of extending via function calls

Implemented over here, and I've added you as a contributor on the commit as it was based on your work
https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/536

PR #453, opened 2021-11-03
prioritize newer mock handlers

dwickern:
When there are multiple mocks for the same request, we should prioritize the newer one. For example, you might set up some mocks for all tests in `beforeEach`, and then override it for a specific test. Currently the first mock would be used.

Besides that, we should put a higher priority to mocks with a filter, regardless of their declaration order. Here's what I came up with:
1. prioritize mocks using `match()` to match the request body
2. prioritize mocks using `withParams()` to match exact query params
3. prioritize mocks using `withSomeParams()` to match partial query params
4. otherwise, prioritize by reverse declaration order

### DIFF
--- a/addon/mocks/mock-create-request.js
+++ b/addon/mocks/mock-create-request.js
@@ -11,7 +11,6 @@ export default class MockCreateRequest extends AttributeMatcher(
     super(modelName, 'createRecord');
     this.model = model;
     this.returnArgs = {};
-    this.matchArgs = {};
     this.setupHandler();
   }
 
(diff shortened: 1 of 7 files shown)


patocallaghan:
Hey @dwickern, this is an interesting idea. I'd be happy to chat more about it. This would be a breaking change so I wonder if we should try and get some feedback from other users of the framework.

Techn1x:
Going to close this one, it's been a few years and I'm not personally invested in the change, and not confident that it would still be neeed.

I would be more than happy to re-open and consider merging if you would indeed find use for it

PR #458, opened 2022-02-12
Change import of deprecate from @ember/application/deprecations to @ember/debug

colinfkennedy:
The existing import is preventing apps upgrading to Ember 4 due to a missing module error. The module `@ember/application/deprecations` no longer exists in Ember 4.

As the `deprecate` function is available in `@ember/debug` in  Ember`v3.28` (Current Ember version for ember-data-factory-guy)  and Ember `v4.0.0` this should be safe to change.

I have tested this locally with my own Ember 4 app.


patocallaghan:
Thanks @colinfkennedy 🙇‍♂️ Have some failing tests but they're unrelated to your changes. Will merge your fixes anyway.

PR #468, opened 2022-08-18
fix: remove hard dependency on model-fragments

luxferresum:
using factory guy without model-fragments results in errors (at least on ember 4.4). this should fix this.

### DIFF
--- a/addon/converter/fixture-converter.js
+++ b/addon/converter/fixture-converter.js
@@ -13,7 +13,9 @@ try {
   Fragment = MF && MF.default;
   FragmentArray = MFA && MFA.default;
 } catch (e) {
-  // do nothing
+  // create empty constructors
+  Fragment = function Fragment() {};
+  FragmentArray = function FragmentArray() {};
 }
 
 /**
--- a/addon/converter/fixture-converter.js
+++ b/addon/converter/fixture-converter.js
@@ -3,8 +3,18 @@ import { entries } from '../utils/helper-functions';
 import { typeOf } from '@ember/utils';
 import { getOwner } from '@ember/application';
 import { camelize } from '@ember/string';
-import Fragment from 'ember-data-model-fragments/fragment';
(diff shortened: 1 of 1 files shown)


GabrielCousin:
Hi @luxferresum and thanks for the fix!
I tried this when upgrade our project to Ember 4.7 and it was not working completely (at least it solved the build but most of the tests were not passing anymore)

```
TypeError: Right-hand side of 'instanceof' is not an object
    at vendor.js:106676:39
    at vendor.js:65308:18
    at Map.forEach (<anonymous>)
    at TaskTemplate.eachAttribute (vendor.js:65307:23)
    at JSONAPIFixtureConverter.extractAttributes (vendor.js:106669:38)
    at JSONAPIFixtureConverter.convertSingle (vendor.js:107092:26)
    at JSONAPIFixtureConverter.convert (vendor.js:106568:21)
    at JSONAPIFixtureBuilder.convertForMake (vendor.js:106315:24)
    at FactoryGuy.make (vendor.js:107634:49)
    at Object.<anonymous> (tests.js:1416:60)
```
TaskTemplate is one our models. I added the following changes here https://github.com/MakeMusicInc/ember-data-factory-guy/commit/30cb1233111534376abfae08b3bfb205f43c3eb1
Do you think it would make sense to add this to that PR?

patocallaghan:
@luxferresum yep we should fix this. If you can apply the fixes @GabrielCousin suggests I'll merge and do a new release

patocallaghan:
Great, thanks I'll follow up this evening 🙇

patocallaghan (changes requested):
@luxferresum looks like you have a lint error. I think we need to import `require` like in `addon/factory-guy.js`

patocallaghan:
Thanks @luxferresum and @GabrielCousin 🙌

patocallaghan:
Released in [3.9.11](https://github.com/adopted-ember-addons/ember-data-factory-guy/releases/tag/v3.9.11)

PR #470, opened 2022-10-20
Make requireFiles do less work for larger applications

misterbyrne:
We have a very large Ember application with loads of modules, so the `requireFiles` function takes a long time.

This is because the `fileSeen` array is very long, but only a very small subset of these match the `filePattern` regex.
The `filePattern` regex is very simple, but the `excludeRegex` is more complex. So if we evaluate the simple one first, we'll only have to evaluate the more complex one a much smaller number of times.

I also took the opportunity to simplify the `excludeRegex` since it's the file extension that we really care about - at the end of the path.

Time spent in `requireFiles` for one test module of our app before the change was 2605ms, time spent after the change was 70ms.

### DIFF
--- a/addon/utils/helper-functions.js
+++ b/addon/utils/helper-functions.js
@@ -206,7 +206,7 @@ export function parseUrl(url) {
 }
 
 // always exclude jshint or jscs files
-export const excludeRegex = new RegExp('[^\\s]+(\\.(jscs|jshint))$', 'i');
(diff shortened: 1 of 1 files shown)


patocallaghan:
Thanks @misterbyrne 🙌

patocallaghan:
Released in [3.9.11](https://github.com/adopted-ember-addons/ember-data-factory-guy/releases/tag/v3.9.11)

Issue #479, opened 2023-01-10
Fix ember-assign deprecation warning


drouhard:
After updating to Ember 4 I am seeing thousands of these deprecation warnings coming from ember-data-factory-guy initializing.

https://github.com/emberjs/rfcs/blob/master/text/0750-deprecate-ember-assign.md

Fortunately the recommended solution is backwards-compatible and simple.
Replace this:
```
import { assign } from '@ember/polyfills';
```
with this:
```
import { assign as emberAssign } from '@ember/polyfills';

const assign = Object.assign || emberAssign;
```

PR #480, opened 2023-01-10
Fix ember-assign deprecation

drouhard:
## Purpose
Address deprecation warnings. (See https://github.com/emberjs/rfcs/blob/master/text/0750-deprecate-ember-assign.md)

Fix https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/479

## Summary
Follow the suggestion in the deprecation RFC and use a backwards-compatible helper for `assign` that pulls in `@ember-polyfill/assign` only when the consumer doesn't have `Object.assign`. (That is, for IE11 users).

### DIFF
--- a/addon/factory-guy.js
+++ b/addon/factory-guy.js
@@ -1,7 +1,7 @@
 import Store from '@ember-data/store';
 import { assert } from '@ember/debug';
 import { isPresent, typeOf } from '@ember/utils';
-import { assign } from '@ember/polyfills';
+import { assign } from './utils/polyfills';
 import { join } from '@ember/runloop';
 import { A } from '@ember/array';
 import require from 'require';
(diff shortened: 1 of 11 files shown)


drouhard:
@patocallaghan I think this should help with https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/479. Are you still the right person to review and get this merged?

patocallaghan:
@drouhard I actually think we can just drop this outright and use `Object.assign` everywhere. How about we just do that instead? We're kind of gearing up for a major version bump and I think that will allow us to make this breaking change (drop IE11 support)

drouhard:
Love it. Closing this PR in favor of taking your suggestion, here: https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/481

PR #486, opened 2023-11-13
Replace `require` with `importSync` to play nice with Embroider

achambers:
This PR removes the use of `require` as a means of deciding whether to import a package or not and replaces with the the @embroider/macros provided `dependencySatisfies` and `importSync`. These functions are backwards compatible and therefore will work in both Classic and Embroider builds.


mansona (approved):
Looks good to me 👍 it would probably be good to get someone more familiar with the code to take a look too but if tests are passing I guess we could just merge it 🤷

PR #487, opened 2023-11-14
Add Embroider ember-try scenarios

achambers:


Techn1x:
Thanks for the PR! the changes here are all present in main branch now, with the embroider scenarios passing - closing

PR #492, opened 2024-08-28
Move ember-inflector to dependencies from devDependencies

drouhard:
Allows consuming apps to upgrade their version of ember-inflector to v5.

If you upgrade ember-inflector to v5 in an app that uses ember-data-factory-guy today, you get this build error:
> ember-data-factory-guy tried to import "ember-inflector" in "ember-data-factory-guy/converter/rest-fixture-converter.js" from addon code, but "ember-inflector" is a devDependency. You may need to move it into dependencies.

See https://github.com/emberjs/ember-inflector/pull/499 for the breaking PR that converted ember-inflector to a v2 addon.


drouhard:
Test failures suggest that this repo needs to upgrade `ember-auto-import` to v2. I see a stale open PR for that here: https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/488

It's unclear what other work needs to be done on that PR.

Techn1x:
Thanks for the PR. I've added this to things for me to look at as I update this repo - I'll likely add it to peerDeps since that's what ember-data seems to do be doing. 

This addon is on ember-auto-import v2 these days so hopefully that's no longer an issue

PR #464, opened 2022-02-15
Remove references to `AdapterFetch`

patocallaghan:
Part of https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/444

Fixes `deprecate-fetch-ember-data-support` deprecations

PR #503, opened 2025-04-29
chore: drop AdapterFetch, ember-fetch, ember-django-adapter

Techn1x:
Have gone a step further than https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/464 in order to deal with the test failures, by dropping ember-django-adapter.

I looked into fixing things for the ember-django-adapter so that it would produce the correct payload/errors without AdapterFetch but it's clear that addon is no longer maintained;
* there's a callout for maintainers from 2022 https://github.com/dustinfarris/ember-django-adapter?tab=readme-ov-file#call-for-maintainers
* there hasn't been any actual code commits in 6 years (2021)
* no issues or PRs in the same amount of time, except my own that never got merged https://github.com/dustinfarris/ember-django-adapter/pull/263

it's likely no-one is using it, so I opted to drop it. If anyone is using it, they'd have some ancient dependencies anyway, I'm sure they wouldn't mind sticking with older versions of factory guy.

Likely a major bump but I plan to do a few more improvements before that.


tests/dummy/app/controllers/users.js
```diff
@@ -1,9 +1,10 @@
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
+import { waitForPromise } from '@ember/test-waiters';
 
 export default class extends Controller {
   @action
   deleteUser(user) {
-    return user.destroyRecord();
+    waitForPromise(user.destroyRecord());
```
Techn1x, line 8: ember-fetch provided a fetch that is registered with settled checks in tests. After dropping it, some tests started failing because they were acting on the records while they were still "inFlight"

So we need to use test-waiters for these, to ensure the test doesn't continue until the requests are settled

config/ember-try.js
```diff
@@ -50,11 +50,6 @@ module.exports = async function () {
...
       {
         name: 'ember-default-with-jquery',
-        env: {
-          EMBER_OPTIONAL_FEATURES: JSON.stringify({
-            'jquery-integration': true,
-          }),
-        },
         npm: {
           devDependencies: {
             '@ember/jquery': '^1.1.0',
```
Techn1x, line 55: After dropping ember-fetch, when `'jquery-integration': true`, tests regarding FormData started failing - specifcally, this assert on this test for each adapter
https://github.com/adopted-ember-addons/ember-data-factory-guy/blob/f1f2a37056f182b367f58247636f2e0c15b3be06/tests/unit/shared-adapter-behaviour.js#L1430-L1436

The FormData just wouldn't come through in any form. I am suspecting that this issue from pretender is kicking in https://github.com/pretenderjs/pretender/issues/159

Either way, `'jquery-integration': true` hasn't been supported in ember for a while now, so I think we are okay to drop support for it. It'll still "work" for people too if they want, as long as they don't use FormData.

I've left the ember-try scenario since it still works fine with `@ember/jquery`

README.md
```diff
@@ -5,32 +5,9 @@
 Feel the thrill and enjoyment of testing when using Factories instead of Fixtures.
 Factories simplify the process of testing, making you more efficient and your tests more readable.
 
-**NEW** starting with v3.8
-  - jquery is no longer required and fetch adapter is used with ember-data
-  - you can still use jquery if you want to  
-  - if you are addon author using factory guy set up your application adapter like [this](https://github.com/adopted-ember-addons/ember-data-factory-guy/blob/master/tests/dummy/app/adapters/application.js)
```
Techn1x, line 11: cleaned up the readme a bit while I was here, and removed django references

PR #504, opened 2025-05-01
chore: add `@ember/string` peerDependency

Techn1x:
This addon uses `@ember/string` in a few places, which is no longer included as part of Ember in later versions. Need it listed in peer deps if using in modern projects/builds with stricter dependency management


Techn1x:
The failing try scenario for ember-release is handled in https://github.com/Techn1x/ember-data-factory-guy/pull/1

PR #505, opened 2025-05-01
refactor: avoid automagic require, use manual imports

Techn1x:
* remove use of `require()` in app-facing code
* remove `useScenarios` flag in favour of app importing and running their own scenarios in tests
* remove automatic import of factories in favour of app importing their own factories in test-helper.js
* update readme


addon/index.js
```diff
@@ -58,8 +47,17 @@ export {
...
 };
 
+/**
+ * Setup and teardown code, intended to be called with qunit hooks so that it can run code before & after each test.
+ */
 export function setupFactoryGuy(hooks) {
   hooks.beforeEach(function () {
-    manualSetup(this);
+    const owner =
+      this.owner || (this.container && this.container.owner) || this;
```
Techn1x, line 56: would have liked to use `getOwner(this)` because simpler, but it doesn't seem available (guessing because ember 3.28? I think it was introduced in 4.x?)

If we end up dropping ember 3.x I'll look into replacing this

addon/index.js
```diff
@@ -58,8 +47,17 @@ export {
...
+    FactoryGuy.setStore(owner.lookup('service:store'));
```
Techn1x, line 57: will look into the store service deprecation stuff later

addon/index.js
```diff
@@ -58,8 +47,17 @@ export {
...
+  });
+
+  hooks.afterEach(function () {
+    FactoryGuy.reset();
```
Techn1x, line 61: Note that we're not calling `loadScenarios()` or `loadFactories()` anymore, avoiding `require()` altogether - it's up to the app to import what they want to use.

A migration guide or the changelog will detail the changes needed, but briefly;
* they'll want to import their factories before any tests run (likely in test-helper.js)
* they'll want to import any scenario(s) they want to use, and call run(), for each test (likely making use of beforeEach hooks)

See readme for more info

---

would have liked to use `getOwner(this)` for the owner code because simpler, but it doesn't seem available (guessing because ember 3.28? I think it was introduced in 4.x?). If we end up dropping ember 3.x I'll look into using it

Will look into the store service deprecations in later work

addon/utils/helper-functions.js
```diff
@@ -203,24 +201,3 @@ export function parseUrl(url) {
...
- */
-export function requireFiles(filePattern) {
-  let filesSeen = Object.keys(requirejs._eak_seen);
-
-  return filesSeen
-    .filter(
-      (moduleName) =>
-        filePattern.test(moduleName) && !excludeRegex.test(moduleName)
-    )
-    .map((moduleName) => require(moduleName, null, null, true));
```
Techn1x, line 225: This here is problematic (I'm currently running an embroider + vite project that uses FG - had to patch this out in my project)

index.js
```diff
@@ -109,14 +89,10 @@ module.exports = {
...
   setupFactoryGuyInclude: function (app) {
     let defaultEnabled = /test|development/.test(app.env),
-      defaultSettings = { enabled: defaultEnabled, useScenarios: false },
+      defaultSettings = { enabled: defaultEnabled },
       userSettings = app.project.config(app.env).factoryGuy || {},
       settings = Object.assign(defaultSettings, userSettings);
 
-    if (settings.useScenarios) {
-      settings.enabled = true;
-    }
```
Techn1x, line 118: `useScenarios` config no longer does anything - if you want to use scenarios, import and run them!

index.js
```diff
@@ -62,25 +61,6 @@ module.exports = {
...
-            destDir: 'tests/factories',
-          });
-          trees.push(factoriesTree);
-        }
-      } catch (err) {
-        // do nothing;
-      }
-    }
-
-    return MergeTrees(trees);
```
Techn1x, line 81: I stepped through this on the main branch, funnily enough the Funnel throws so the catch block is always hit - this function doesn't do anything, deleted

tests/test-helper.js
```diff
@@ -5,6 +5,8 @@ import { setApplication } from '@ember/test-helpers';
 import { setup } from 'qunit-dom';
 import { start } from 'ember-qunit';
 
+import 'dummy/tests/factories';
+
 setApplication(Application.create(config.APP));
```
Techn1x, line 10: An example of importing factories in the test-helper.js file

addon/index.js
```diff
@@ -58,8 +48,16 @@ export {
...
+ */
 export function setupFactoryGuy(hooks) {
   hooks.beforeEach(function () {
-    manualSetup(this);
+    const { owner } = getContext();
+    FactoryGuy.setStore(owner.lookup('service:store'));
+  });
+
+  hooks.afterEach(function () {
+    FactoryGuy.reset();
```
Techn1x, line 61: Note that we're not calling `loadScenarios()` or `loadFactories()` anymore, avoiding `require()` altogether - it's up to the app to import what they want to use.

A migration guide or the changelog will detail the changes needed, but briefly;
* they'll want to import their factories before any tests run (likely in test-helper.js)
* they'll want to import any scenario(s) they want to use, and call run(), for each test (likely making use of beforeEach hooks)

See readme for more info

---

Will look into the store service deprecations in later work

Techn1x:
The failing try scenario for ember-release is handled in https://github.com/Techn1x/ember-data-factory-guy/pull/1

PR #507, opened 2025-05-02
test: tests working with ember 4.12 and ember-data 3.28 + embroider try scenarios

Techn1x:
see review comments


package.json
```diff
@@ -49,19 +46,19 @@
...
     "broccoli-asset-rev": "^3.0.0",
     "ember-cli": "~3.28.0",
     "ember-cli-dependency-checker": "^3.2.0",
-    "ember-cli-deprecation-workflow": "^1.0.1",
+    "ember-cli-deprecation-workflow": "^3.3.0",
     "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-terser": "^4.0.2",
-    "ember-data": "~3.24",
+    "ember-data": "~3.28.13",
```
Techn1x, line 61: Needed to avoid use of Ember global - later ember versions remove the ember global

package.json
```diff
@@ -49,19 +46,19 @@
...
+    "@embroider/test-setup": "^4.0.0",
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
     "active-model-adapter": "^3.0.1",
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
     "ember-cli": "~3.28.0",
     "ember-cli-dependency-checker": "^3.2.0",
-    "ember-cli-deprecation-workflow": "^1.0.1",
+    "ember-cli-deprecation-workflow": "^3.3.0",
```
Techn1x, line 57: old version also uses ember global, so need to update it

config/ember-try.js
```diff
@@ -1,29 +1,23 @@
...
-        npm: {
-          devDependencies: {
-            'ember-source': '~3.24.3',
+            'ember-source': '~4.12',
           },
         },
       },
       {
         name: 'ember-release',
+        allowedToFail: true,
```
Techn1x, line 20: Will enforce this later, after fixing and adding ember 5.12 try scenario. Need to update a lot of deps before ember 6 try scenario will work

tests/dummy/app/routes/cats.js
```diff
@@ -1,6 +1,9 @@
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
 
 export default class extends Route {
+  @service store;
+
   model() {
     return this.store.findAll('cat');
```
Techn1x, line 8: later ember versions don't have implicit service or store injections

tests/dummy/app/routes/employee.js
```diff
@@ -0,0 +1,12 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class extends Route {
+  @service store;
+
+  model(params) {
+    return this.store
+      .findRecord('employee', params.employee_id)
+      .catch(() => null);
```
Techn1x, line 10: later ember versions don't have implicit model hook that makes use of ember data store, so write it explicitly

tests/unit/shared-adapter-behaviour.js
```diff
@@ -1654,15 +1654,21 @@ SharedBehavior.mockUpdateWithErrorMessages = function () {
 
       mockUpdate(profile).fails({
         status: 400,
-        response: { errors: { description: ['invalid data'] } },
+        /* ember-data 3.28 -> 4.x rest adapter normalizeErrorResponse only passes on string response value as 'detail'
+         * ember-data 5.0 fixes this, but we cannot use it yet https://github.com/emberjs/data/pull/8621
+         * Just use a primitive value in the meantime.
+         */
+        // response: { errors: { description: ['invalid data'] } },
+        response: 'invalid data',
```
Techn1x, line 1662: ember-data-model-fragments is holding us back from ember-data 5 which contains the fix we want for this to keep working. This is a pragmatic solution to keep the bulk of the test and move forward

---

If EDMF holds us back in more ways, it might be worth considering dropping it as well - from what I can tell, ember-data 4.12+ has built-in support for fragments and so the addon is likely to be deprecated/archived soon in favour of users refactoring to that approach.
https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/471#issuecomment-1637407820

On brief inspection of compatibility, I think we might have to drop it. If we want to test against ember-source 5.12 (and higher) ember-data 4.12 is the min version supporting that. EDMF doesn't support anything above ember-data 4.6

PR #510, opened 2025-05-02
refactor: convert to v2 addon, drop node <= 16

Techn1x:
Followed this guide https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md

Notes:
* v2 addons require the consuming app to have ember-auto-import v2, so this will be marked as a breaking change for that reason
* needed to drop node <= 16 to get CI building with rollup. So breaking change for that reason too.
* removed the `settings.enabled` and dev/test/prod checks for tree build exclusion that index.js was doing before
  * in theory, it should all tree-shake out if all FG uses are in test code. There should also be no more need for settings.enabled config. If the user has deeper requirements of when to shake it out of the build, they can use embroider/macros importSync() and isDevelopingApp() + isTestingApp()


addon/addon/builder/fixture-builder-factory.js
```diff
@@ -5,17 +5,15 @@ import JSONAPIFixtureBuilder from './jsonapi-fixture-builder';
...
+import { importSync } from '@embroider/macros';
 
 let ActiveModelSerializer;
-if (macroCondition(dependencySatisfies('active-model-adapter', '*'))) {
+try {
   ActiveModelSerializer = importSync(
     'active-model-adapter'
   ).ActiveModelSerializer;
+} catch (e) {
+  // no worries, go on without it - it's optional
```
Techn1x, line 16: Note this change. `dependencySatisfies()` doesn't work for optional peerdeps (will only work if the dep is listed as dependency, which we don't want) - rather, we can just use try/catch. If they don't have the package, that's OK.

Made the same change for the ember-data-model-fragment imports.

Some references
https://github.com/embroider-build/embroider/pull/1187/files#r855292577
https://github.com/embroider-build/embroider/issues/1266

addon/addon/converter/fixture-converter.js
```diff
@@ -3,19 +3,17 @@ import { entries } from '../utils/helper-functions';
...
 let Fragment;
 let FragmentArray;
-if (macroCondition(dependencySatisfies('ember-data-model-fragments', '*'))) {
+try {
   Fragment = importSync('ember-data-model-fragments/fragment').default;
   FragmentArray = importSync(
     'ember-data-model-fragments/array/fragment'
   ).default;
+} catch (e) {
+  // no worries, go on without it - it's optional
```
Techn1x, line 16: see https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/510/files#r2071274861

addon/addon/factory-guy.js
```diff
@@ -6,17 +6,15 @@ import { A } from '@ember/array';
...
+import { importSync } from '@embroider/macros';
 
 let modelDefinitions = {};
 
 let Fragment;
-if (macroCondition(dependencySatisfies('ember-data-model-fragments', '*'))) {
+try {
   Fragment = importSync('ember-data-model-fragments').default.Fragment;
+} catch (e) {
+  // no worries, go on without it - it's optional
```
Techn1x, line 17: see https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/510/files#r2071274861

index.js
```diff
@@ -1,117 +0,0 @@
...
-    // from being added to app tree. Would have thought that this would have
-    // happened in treeForApp above, but not the case
-    if (
-      !this.includeFactoryGuyFiles &&
-      this.treeExcludeRegex &&
-      this.treeExcludeRegex.test(name)
-    ) {
-      return;
-    }
-    return this._super.treeFor.apply(this, arguments);
```
Techn1x, line 115: TODO: implement this with embroider/macros

The rest of the pretender package shim stuff in this file shouldn't be necessary

addon/src/builder/fixture-builder-factory.js
```diff
@@ -5,17 +5,15 @@ import JSONAPIFixtureBuilder from './jsonapi-fixture-builder';
...
+import { importSync } from '@embroider/macros';
 
 let ActiveModelSerializer;
-if (macroCondition(dependencySatisfies('active-model-adapter', '*'))) {
+try {
   ActiveModelSerializer = importSync(
     'active-model-adapter'
   ).ActiveModelSerializer;
+} catch (e) {
+  // no worries, go on without it - it's optional
```
Techn1x, line 16: Note this change. `dependencySatisfies()` doesn't work for optional peerdeps (will only work if the dep is listed as dependency, which we don't want) - rather, we can just use try/catch. If they don't have the package, that's OK.

Made the same change for the ember-data-model-fragment imports.

Some references
https://github.com/embroider-build/embroider/pull/1187/files#r855292577
https://github.com/embroider-build/embroider/issues/1266

addon/src/converter/fixture-converter.js
```diff
@@ -3,19 +3,17 @@ import { entries } from '../utils/helper-functions';
...
 let Fragment;
 let FragmentArray;
-if (macroCondition(dependencySatisfies('ember-data-model-fragments', '*'))) {
+try {
   Fragment = importSync('ember-data-model-fragments/fragment').default;
   FragmentArray = importSync(
     'ember-data-model-fragments/array/fragment'
   ).default;
+} catch (e) {
+  // no worries, go on without it - it's optional
```
Techn1x, line 16: see https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/510/files#r2074747452

addon/src/factory-guy.js
```diff
@@ -6,17 +6,15 @@ import { A } from '@ember/array';
...
+import { importSync } from '@embroider/macros';
 
 let modelDefinitions = {};
 
 let Fragment;
-if (macroCondition(dependencySatisfies('ember-data-model-fragments', '*'))) {
+try {
   Fragment = importSync('ember-data-model-fragments').default.Fragment;
+} catch (e) {
+  // no worries, go on without it - it's optional
```
Techn1x, line 17: see https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/510/files#r2074747452

Issue #485, opened 2023-07-05
String vs number ids and ember-data 5.3+ issues


Techn1x:
I'd like to use string ids across the board, as that's what my JSONAPI does and it's also what ember-data generally expects.

EDFG seems to always use number ids by default when using a factory.
https://github.com/adopted-ember-addons/ember-data-factory-guy/blob/c4ccbaa70880084c13097c8c4ab6e4e1d6f5434b/addon/model-definition.js#L80-L82

That might have worked in the past, but I think ember-data is a bit more strict now. On ember-data 4.12.1, it seems that for a patch/update payload, if ED receives an id from the api (string) that does match match what is cached in its store (number), it fails with this error;

```
Error: Assertion Failed: Expected the ID received for the primary 'teacher' resource being saved to match the current id '1' but received '1'.
    at assert (index.js:153:1)
    at JSONAPICache.didCommit (index.js:627:103)
    at eval (index.js:432:22)
    at Store._run (index-757cf686.js:5403:5)
    at Store._join (index-757cf686.js:5419:12)
    at eval (index.js:429:11)
```

This can be replicated with code like the following;
```js
const teach = make('teacher') // pushes a teacher to the store with id 1 (number)
// console.log(teach.id) // record in store has id '1' (string)
mockUpdate(this.teacher)
  .match({ locale: 'fr' })
  .returns({ attrs: { language: 'fr' } })
teach.name = "newname"
await teach.save()
```

---

The workaround is to set an id in the model when creating it `make('teacher', { id: 'my-string-id' })` but I've got a lot of tests to update if I go that path..

I think the fix is to either force string ids from that nextId function, or give an option to users to set a flag in a factory to use string ids.

Techn1x:
Looking deeper, it might be specifically something to do with the `mockUpdate().match(...).returns(...)`

Either way, I still think giving the option to use string ids in factories would be beneficial

Techn1x:
Also, I think string ids is be a requirement of ember-data 5 & 6.

It's a deprecation item here until ED6
https://deprecations.emberjs.com/ember-data/v5.x#toc_ember-data-deprecate-non-strict-id
> We are deprecating this legacy support for numeric IDs.

But even when using ED with `compatWith: '5.3'` it will fail with
```
Error: Assertion Failed: Resource IDs must be a non-empty string or null. Received '1'.
```

esbanarango:
Plus 1

PR #521, opened 2025-05-08
test: replace runloop usage, use assert.strictEqual

Techn1x:
Mostly odds and ends
* don't need to use runloop in tests
* use assert.strictEqual
* added some handy pnpm scripts
* correct name for FixtureBuilder import


README.md
```diff
@@ -1706,7 +1703,6 @@ Usage:
        json, returning `true` if there is a match, `false` otherwise.
   - returns
     - attributes ( including relationships ) to include in response json
-- Need to import `run` from `@ember/runloop` and wrap tests using `mockCreate` with: `run(function() { 'your test' })`
```
Techn1x, line 1709: As far as I can tell, this isn't needed in tests anymore. I suspect testing has improved a lot since this was written

addon/package.json
```diff
@@ -24,7 +24,10 @@
   "scripts": {
     "build": "rollup --config",
     "prepublishOnly": "rollup --config",
-    "start": "rollup --config --watch"
+    "start": "rollup --config --watch",
+    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
+    "lint:js": "pnpm -w eslint ./addon",
+    "lint:js:fix": "pnpm -w eslint ./addon --fix"
```
Techn1x, line 30: Run eslint, from the workspace root, so that it uses the monorepo config defined there.

test-app/tests/unit/json-adapter-test.js
```diff
@@ -29,37 +28,30 @@ module(serializer, function (hooks) {
...
-          });
-      });
+    test('when returns json (plain) is used', async function (assert) {
+      let json = { id: 1, description: 'the desc' },
+        mock = mockFindRecord('profile').returns({ json }),
+        profileId = mock.get('id');
+
+      const profile = await FactoryGuy.store.findRecord('profile', profileId);
+
+      assert.strictEqual(profile.get('id'), profileId.toString());
```
Techn1x, line 38: `get.id('id')` from a store record produces a string, whereas the id returned by FG is (currently) a number. strictEqual brings this to light, so added toString() where this is relevant.

We'll want to later refactor FG to be strings only. related https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/485

Issue #493, opened 2024-09-13
store.willDestroy error when using ember-data with DEPRECATE_STORE_EXTENDS_EMBER_OBJECT


Techn1x:
Using ember-data 5.3.4+ with `emberData: { compatWith: '5.2' },`

I seem to hit this error;

![Screenshot 2024-09-13 at 10 39 10 am](https://github.com/user-attachments/assets/bae1dffa-f82b-485a-9c7a-7f8d2eba0953)

Techn1x:
Pasting the deprecation here for later
```
DEPRECATION: The Store class extending from EmberObject is deprecated.
Please remove usage of EmberObject APIs and mark your class as not requiring it.

To mark the class as no longer extending from EmberObject, in ember-cli-build.js
set the following config:

```js
const app = new EmberApp(defaults, {
  emberData: {
    deprecations: {
      DEPRECATE_STORE_EXTENDS_EMBER_OBJECT: false
    }
  }
});
```
 [deprecation id: ember-data:deprecate-store-extends-ember-object] This will be removed in ember-data 6.0
```

Techn1x:
Resolved by https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/522

PR #522, opened 2025-05-08
refactor: deal with using Store EmberObject function deprecations

Techn1x:
Resolves https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/493

Ensures we don't expect `store.willDestroy()` to exist, as Store is not an EmberObject in upcoming versions of ember-data. This deals with the deprecation, and have configured the test-app to error if it tries to do so.

I also configured the tests to have better test isolation, and dealt with violations of that.

### DIFF
--- a/addon/src/factory-guy.js
+++ b/addon/src/factory-guy.js
@@ -41,7 +41,6 @@ class FactoryGuy {
     );
     this.store = aStore;
     this.fixtureBuilderFactory = new FixtureBuilderFactory(this.store);
-    this.afterDestroyStore(aStore);
   }
 
   fixtureBuilder(modelName) {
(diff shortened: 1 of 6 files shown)


addon/src/factory-guy.js
```diff
@@ -489,6 +488,8 @@ class FactoryGuy {
   }
 
   reset() {
+    this.store = null;
+    this.fixtureBuilderFactory = null;
     this.resetDefinitions();
     this.resetMockAjax();
```
Techn1x, line 494: Rather than hooking into store.willDestroy to clear the store - we instead clear the store on each reset() call. This seems to work fine.

PR #525, opened 2025-05-09
chore: more ember-try scenarios, min ember/test-helpers now 2.9.3+

Techn1x:


test-app/config/ember-try.js
```diff
@@ -20,6 +20,24 @@ module.exports = async function () {
         npm: {
           devDependencies: {
             'ember-source': '~5.12',
+            'ember-load-initializers': '^3.0.0', // v3 needed for ember 5+
+            'ember-data': '~5.3', // couldnt get earlier versions to work with ember-source 5.12
```
Techn1x, line 24: technically ember-data 4.12 should support ember-source 5.x but I could not get past this error until I bumped ember-data to 5.x
```
Global error: Uncaught TypeError: Cannot set property default of #<Object> which has only a getter
```

addon/package.json
```diff
@@ -65,7 +65,7 @@
     "@ember-data/serializer": ">= 3.28",
     "@ember-data/store": ">= 3.28",
     "@ember/string": ">= 3.0.0",
-    "@ember/test-helpers": ">= 2.4.2",
+    "@ember/test-helpers": "^2.9.3 || >= 3",
```
Techn1x, line 68: This could be considered breaking, but I don't think this will be an issue for people.

Latest v2 supports later versions of ember etc etc and has lots of backported fixes.

.github/workflows/ci.yml
```diff
@@ -76,6 +76,10 @@ jobs:
           ]
         allow-failure: [false]
         include:
+          - ember-try-scenario: ember-lts-5.12
+            allow-failure: true
+          - ember-try-scenario: ember-6.4
+            allow-failure: true
```
Techn1x, line 82: These are failing at the moment, but it's likely failing because of issues with the tests themselves. Will fix them up later.

I have a project on ED v5.3 + ember-source 6.4 that uses factory guy, so I am confident the addon itself is OK for scenarios.

PR #527, opened 2025-05-09
chore: bump test-app devdeps, drop ember-sinon

Techn1x:
Just bumping test-app devDependencies to latest possible that aren't peerdep contracts (where for those we need to test lower versions)

Updating ember-try scenarios with expected peerdeps for those ember-source versions as I go

Drop ember-sinon / ember-sinon-qunit, we can do what it does with minimal code. Also remove unneeded sinon restore() calls


addon/package.json
```diff
@@ -66,7 +66,7 @@
     "@ember-data/store": ">= 3.28",
     "@ember/string": ">= 3.0.0",
     "@ember/test-helpers": "^2.9.3 || >= 3",
-    "active-model-adapter": "*",
+    "active-model-adapter": "^3.0.1 || ^4.0.0",
```
Techn1x, line 69: these are the only (currently) released versions

test-app/package.json
```diff
@@ -33,19 +33,19 @@
...
+    "@glimmer/component": "^2.0.0",
+    "@glimmer/tracking": "^1.1.2",
     "active-model-adapter": "^3.0.1",
     "ember-auto-import": "^2.10.0",
-    "ember-cli": "~3.28.0",
-    "ember-cli-babel": "^7.26.6",
-    "ember-cli-htmlbars": "^5.7.1",
+    "ember-cli": "~4.12.0",
+    "ember-cli-babel": "^8.2.0",
+    "ember-cli-htmlbars": "^6.3.0",
```
Techn1x, line 48: just bumping stuff to latest that aren't peerdep contracts (where for those we need to test lower versions)

test-app/tests/helpers/install-sinon.js
```diff
@@ -0,0 +1,10 @@
+import sinon from 'sinon';
+
+/**
+ * Ensures after each test that sinon is restored, avoiding stubs/mocks being shared between tests
+ */
+export function installSinon(QUnit) {
+  QUnit.testDone(function () {
+    sinon.restore();
+  });
```
Techn1x, line 9: Pretty much a drop in replacement for ember-sinon / ember-sinon-qunit

PR #530, opened 2025-05-12
feat: added useStringIdsOnly config option to avoid deprecated behaviour

Techn1x:
Resolves https://github.com/adopted-ember-addons/ember-data-factory-guy/issues/485

### DIFF
--- a/addon/src/mocks/attribute-matcher.js
+++ b/addon/src/mocks/attribute-matcher.js
@@ -2,6 +2,7 @@ import { assert } from '@ember/debug';
 import { typeOf } from '@ember/utils';
 import FactoryGuy from '../factory-guy';
 import { isEmptyObject, isEquivalent } from '../utils/helper-functions';
+import { verifyId } from '../own-config';
 
 /**
  This is tricky, but the main idea here is:
@@ -87,6 +88,8 @@ const AttributeMatcher = (superclass) =>
       Valid keys are ${validReturnsKeys}. You used these invalid keys: ${invalidKeys}`,
         invalidKeys.length === 0,
       );
+
+      if (options.attrs?.id) verifyId(options.attrs.id);
 
       return responseKeys;
     }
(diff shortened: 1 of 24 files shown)


Techn1x:
Test all passing when config is not enabled yet, good sign of no regression

addon/src/mocks/attribute-matcher.js
```diff
@@ -88,6 +89,8 @@ const AttributeMatcher = (superclass) =>
         invalidKeys.length === 0,
       );
 
+      if (options.attrs?.id) verifyId(options.attrs.id);
```
Techn1x, line 92: Anywhere the user can provide an `id` to factory guy, check it

addon/src/model-definition.js
```diff
@@ -4,6 +4,27 @@ import MissingSequenceError from './missing-sequence-error';
...
+  /**
+   * An unused id is requested. Return the unused id, and cycle it to make a new one. For strings, we can just return a
+   * unique uuid. For ints, we have to increment to the next id, as the only way to ensure it is somewhat-unique for
+   * each record created.
+   */
+  nextId() {
+    const { newId } = this;
+    this.newId = typeof newId === 'string' ? uuid() : newId + 1;
+    return newId;
+  }
```
Techn1x, line 26: Expecting to drop number ids in next major or so. When that's done, there won't be a need for this class, can just use `uuid()` directly all the time

addon/src/model-definition.js
```diff
@@ -127,10 +138,7 @@ class ModelDefinition {
 
     // set the id, unless it was already set in opts
     if (!fixture.id && !opts.id) {
-      // Setting a flag to indicate that this is a generated an id,
-      // so it can be rolled back if the fixture throws an error.
-      fixture._generatedId = true;
```
Techn1x, line 132: I don't think it's necessary to "rollback an id" if the make failed - and isn't something we can do with string ids. It a make failed, then that generated id just won't get used

PR #532, opened 2025-05-13
chore: drop ember-source 3.28 support (min v4.12+), bump active-model-adapter (min v4+), bump ember-data-model-fragments (min v6+)

Techn1x:
big title so it's all in the changelog

This doesn't necessarily mean this won't run with ember 3.28, just that we're not explicitly testing it. Noting that the ember/test-helpers v3 peerDep does not support ember 3.x


addon/package.json
```diff
@@ -65,9 +65,9 @@
     "@ember-data/serializer": ">= 3.28",
     "@ember-data/store": ">= 3.28",
     "@ember/string": ">= 3.0.0",
-    "@ember/test-helpers": "^2.9.3 || >= 3",
-    "active-model-adapter": "^3.0.1 || ^4.0.0",
-    "ember-data-model-fragments": "^5.0.0 || ^6.0.10",
+    "@ember/test-helpers": ">= 3",
+    "active-model-adapter": "^4.0.0",
+    "ember-data-model-fragments": "^6.0.10",
```
Techn1x, line 70: In theory, no-one should have been on ember-data-model-fragments v5 since min ember-data is 3.28 and that is onlyt supported by EDMF v6+

So updating EDMF peer is mostly ceremony

test-app/package.json
```diff
@@ -52,9 +52,9 @@
     "ember-data-model-fragments": "^6.0.10",
     "ember-inflector": "^6.0.0",
     "ember-load-initializers": "^2.1.2",
-    "ember-qunit": "^6.2.0",
-    "ember-resolver": "^8.1.0",
-    "ember-source": "~3.28.0",
+    "ember-qunit": "^9.0.2",
+    "ember-resolver": "^13.1.0",
+    "ember-source": "~4.12.0",
```
Techn1x, line 57: unblocked from using later versions of ember-qunit, ember-resolver ,etc

test-app/tests/test-helper.js
```diff
@@ -14,6 +15,8 @@ setApplication(Application.create(config.APP));
 
 setup(QUnit.assert);
 
+setupEmberOnerrorValidation();
+loadTests();
```
Techn1x, line 19: requirement of ember-qunit v9

PR #536, opened 2025-05-14
feat: allow `match()` on any mock request

Techn1x:
* Moved the attribute matcher code further down so that it is shared, avoiding a mixin-like pattern with AttributeMatcher
* avoid another legacy mixin-like pattern with MaybeIdOrUrl

heavily based off of @dwickern's work in https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/452


addon/src/mocks/maybe-id-url-match.js
```diff
@@ -1,34 +0,0 @@
...
-   Used by getUrl to => this.get('id')
-
-   For mockDelete: If the id is null, the url will not include the id, and
-   can therefore be used to match any delete for this modelName
-   */
-    get(...args) {
-      if (args[0] === 'id') {
-        return this.id;
-      }
-    }
```
Techn1x, line 19: didn't need to copy this over - classes making use of MaybeIdUrlMatch were already extended from a class that provided this function exacftly

addon/src/mocks/maybe-id-url-match.js
```diff
@@ -1,34 +0,0 @@
...
-     *
-     * @returns {String} url
-     */
-    getUrl() {
-      let url = super.getUrl();
-      if (!this.id) {
-        url = `${url}/:id`;
-      }
-      return url;
-    }
```
Techn1x, line 31: copied into mock-delete-request and mock-update-request

PR #540, opened 2025-05-14
chore: run tests against ember-data 5.3, explicitly support ember 4.12, 5.12, 6.4

Techn1x:
* Drop explicit support for ember-data < 5.3 in terms of what we're testing for
* ember-source 5.12 and 6.4 try scenarios are now required to pass CI before PRs can be merged

TLDR; Now explicitly supporting;
* ember-source 4.12, 5.12, 6.4
* ember-data 5.3+

Versions outside of that matrix can't be guaranteed to work. They _probably_ will, but we're not testing for it. I've kept the addon's peerDependency for ED at 3.28 _for now_ since it will likely continue to work in your application. But this addon should reserve the right to drop that if it makes things difficult or holds the addon back.

There was a lot of churn between ED 3.28 and 5.3 (eg proxy changes), it would be too challenging to support all of those in an addon with this smaller userbase compared to other addons.

---

_Note that this doesn't mean FG supports ember-data's new features like RequestManager. Just that this gets FG to a version where it would be possible to implement that new stuff - whether that's in this addon or your own application._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,13 +67,15 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          [ember-default-with-jquery, embroider-safe, embroider-optimized]
+          [
+            ember-lts-5.12,
+            ember-6.4,
+            ember-default-with-jquery,
+            embroider-safe,
+            embroider-optimized,
+          ]
         allow-failure: [false]
         include:
-          - ember-try-scenario: ember-lts-5.12
-            allow-failure: true
-          - ember-try-scenario: ember-6.4
-            allow-failure: true
           - ember-try-scenario: ember-release
             allow-failure: true
 
--- a/addon/src/converter/fixture-converter.js
+++ b/addon/src/converter/fixture-converter.js
@@ -96,11 +96,13 @@ export default class FixtureConverter {
     }
   }
 
-  transformRelationshipKey(relationship) {
-    let { parentModelName, parentType } = relationship,
-      type =
+  transformRelationshipKey(relationship, parentModelName) {
+    const type =
+        /* parentModelName and parentType are not present on relationship schema in ember-data 5+, but these are here for backwards compat */
+        relationship.parentModelName ||
+        relationship.parentType?.modelName ||
+        /**/
         parentModelName ||
-        (parentType && parentType.modelName) ||
         relationship.type,
       transformFn = this.getTransformKeyFunction(type, 'Relationship');
     return transformFn(relationship.name, relationship.kind);
@@ -246,7 +248,7 @@ export default class FixtureConverter {
       ),
       relationshipKey = isEmbedded
         ? relationship.name
-        : this.transformRelationshipKey(relationship);
+        : this.transformRelationshipKey(relationship, parentModelName);
 
     let data = this.extractSingleRecord(
       belongsToRecord,
@@ -275,7 +277,10 @@ export default class FixtureConverter {
 
   extractHasMany(fixture, relationship, parentModelName, relationships) {
     let hasManyRecords = fixture[relationship.name],
-      relationshipKey = this.transformRelationshipKey(relationship),
+      relationshipKey = this.transformRelationshipKey(
+        relationship,
+        parentModelName,
+      ),
       isEmbedded = this.isEmbeddedRelationship(
         parentModelName,
         relationship.name,
@@ -287,6 +292,7 @@ export default class FixtureConverter {
         relationship,
         relationships,
         isEmbedded,
+        parentModelName,
       );
     }
 
@@ -382,8 +388,17 @@ export default class FixtureConverter {
   }
 
   // listProxy data is data that was build with FactoryGuy.buildList method
-  addListProxyData(jsonProxy, relationship, relationships, isEmbedded) {
-    let relationshipKey = this.transformRelationshipKey(relationship);
+  addListProxyData(
+    jsonProxy,
+    relationship,
+    relationships,
+    isEmbedded,
+    parentModelName,
+  ) {
+    let relationshipKey = this.transformRelationshipKey(
+      relationship,
+      parentModelName,
+    );
 
     let records = jsonProxy.getModelPayload().map((data) => {
       if (isEmbedded) {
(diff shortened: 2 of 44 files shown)


.github/workflows/ci.yml
```diff
@@ -67,13 +67,15 @@ jobs:
...
+            embroider-safe,
+            embroider-optimized,
+          ]
         allow-failure: [false]
         include:
-          - ember-try-scenario: ember-lts-5.12
-            allow-failure: true
-          - ember-try-scenario: ember-6.4
-            allow-failure: true
           - ember-try-scenario: ember-release
```
Techn1x, line 79: 🎉

addon/src/converter/fixture-converter.js
```diff
@@ -96,11 +96,13 @@ export default class FixtureConverter {
...
+  transformRelationshipKey(relationship, parentModelName) {
+    const type =
+        /* parentModelName and parentType are not present on relationship schema in ember-data 5+, but these are here for backwards compat */
+        relationship.parentModelName ||
+        relationship.parentType?.modelName ||
+        /**/
         parentModelName ||
-        (parentType && parentType.modelName) ||
         relationship.type,
       transformFn = this.getTransformKeyFunction(type, 'Relationship');
```
Techn1x, line 107: As the comment mentions. This was the one non-test change I had to make in supporting ember-data 5.3. Logic should be exactly the same for ED 3.28 users, and ED 5.x users will get the parentModelName from the args (which should be identical to what was being used before)

patches/active-model-adapter.patch
```diff
@@ -0,0 +1,67 @@
...
+   }
+ }
++
++/**
++ * Adding errorsHashToArray() func until next version of active-model-adapter is released, this function is removed in
++ * ember-data 5+
++ *
++ * https://github.com/adopted-ember-addons/active-model-adapter/issues/191
++ * https://github.com/adopted-ember-addons/active-model-adapter/pull/192
++ */
```
Techn1x, line 26: blerg

ember-data removed a function that active-model-adapter relies on. There's a PR to add it back, which was merged, but a new version of that addon has not been published yet.

Patching it in for now. This will only affect our test-app, to make sure the tests pass for the ED 5 + ActiveModel scenarios. Users on ED 5 + ActiveModel will have to do the same (regardless of whether they use factory guy addon or not) so we can assume they have done so.

test-app/app/models/big-hat.js
```diff
@@ -2,5 +2,5 @@ import Hat from './hat';
 import { hasMany } from '@ember-data/model';
 
 export default class extends Hat {
-  @hasMany('soft-material', { async: false }) materials;
+  @hasMany('soft-material', { async: false, inverse: 'hat' }) materials;
```
Techn1x, line 5: ED 5 requires all async/inverse/as relationship options to be explicitly defined - no more defaults and assumptions

test-app/config/ember-try.js
```diff
@@ -14,8 +14,6 @@ module.exports = async function () {
           devDependencies: {
             'ember-source': '~5.12',
             'ember-load-initializers': '^3.0.0', // v3 needed for ember 5+
-            'ember-data': '~5.3', // couldnt get earlier versions to work with ember-source 5.12
-            'ember-inflector': '^6.0.0', // higher ED version needs this
```
Techn1x, line 18: very glad the try scenarios are getting simpler

test-app/package.json
```diff
@@ -47,7 +47,7 @@
     "ember-cli-babel": "^8.2.0",
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-inject-live-reload": "^2.1.0",
-    "ember-data": "~3.28.13",
+    "ember-data": "~5.3",
```
Techn1x, line 50: test-app now running tests against 5.3

If it was easy as adding a try scenario for ED 3 and 4.12 I would, but the tests need to be different for each, too many things changed

test-app/tests/acceptance/profiles-view-test.js
```diff
@@ -1,15 +1,12 @@
...
-  setupFactoryGuy,
-} from 'ember-data-factory-guy';
+import FactoryGuy, { makeList, mockFindAll } from 'ember-data-factory-guy';
 import { visit } from '@ember/test-helpers';
+import { inlineSetup } from '../helpers/utility-methods';
 
 module('Acceptance | Profiles View', function (hooks) {
   setupApplicationTest(hooks);
-  setupFactoryGuy(hooks);
+  inlineSetup(hooks, '-json-api');
```
Techn1x, line 9: We deleted the default application adapter/serializer for the test-app (in a previous PR). These were JSONAPI.

In 5.3 it no longer provides a default application adapter/serializer for models, so gotta define one, so we use the inline setup to do so

test-app/tests/unit/active-model-adapter-test.js
```diff
@@ -232,12 +233,11 @@ module('DS.ActiveModelSerializer', function (hooks) {
...
-
-      let savedKeyForAttributeFn = serializer.keyForAttribute;
-      serializer.keyForAttribute = dasherize;
-      let savedKeyForRelationshipFn = serializer.keyForRelationship;
-      serializer.keyForRelationship = dasherize;
+      class TestProfileSerializer extends ActiveModelSerializer {
+        keyForAttribute = dasherize;
+        keyForRelationship = dasherize;
+      }
+      this.owner.register('serializer:profile', TestProfileSerializer);
```
Techn1x, line 240: 100% nicer

test-app/tests/unit/rest-adapter-test.js
```diff
@@ -67,8 +67,11 @@ module('DS.RESTSerializer', function (hooks) {
         .createRecord('entry-type', { entries: [entry] })
         .save();
 
-      let entries = entryType.get('entries');
-      assert.deepEqual(entries.mapBy('id'), [entry.id]);
+      let entries = await entryType.entries;
+      assert.deepEqual(
+        entries.map((e) => e.id),
+        [entry.id],
```
Techn1x, line 73: no more mapBy

test-app/tests/unit/shared-adapter-behaviour.js
```diff
@@ -237,8 +236,8 @@ SharedBehavior.mockFindRecordSideloadingTests = function () {
       mockFindRecord('user').returns({ json });
 
       const user = await FactoryGuy.store.findRecord('user', json.get('id'));
-      assert.ok(user.get('hats.firstObject.id') === hat1.get('id') + '');
-      assert.ok(user.get('hats.lastObject.id') === hat2.get('id') + '');
+      assert.ok(user.hats[0].id === hat1.get('id') + '');
+      assert.ok(user.hats[user.hats.length - 1].id === hat2.get('id') + '');
```
Techn1x, line 240: no more firstObject/lastObject

Techn1x:
ugh merged this into wrong branch. reopened here targetting master https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/542

PR #543, opened 2025-05-15
test: GJS for the test templates and components

Techn1x:
* gjs for test templating
* deal with some deprecations in tests
* add try scenario for ember-data 5.5


test-app/ember-cli-build.js
```diff
@@ -16,5 +11,12 @@ module.exports = function (defaults) {
     },
   });
 
+  const { setConfig } = await import('@warp-drive/build-config');
+  setConfig(app, __dirname, {
+    deprecations: {
+      DEPRECATE_STORE_EXTENDS_EMBER_OBJECT: false, // error if using store emberobject methods
+    },
+  });
```
Techn1x, line 19: unrelated: use the new method for providing ember-data with config

test-app/app/transforms/boolean.js
Techn1x, line 1: explicit transform exports, for deprecations that come in later versions of ED

PR #545, opened 2025-05-16
chore: pnpm 10.11, enable ED 5.5 try scenario

Techn1x:


test-app/config/ember-try.js
```diff
@@ -6,7 +6,8 @@ const { embroiderSafe, embroiderOptimized } = require('@embroider/test-setup');
 module.exports = async function () {
   return {
     packageManager: 'pnpm',
-    npmOptions: ['--config.strict-peer-dependencies=false'],
+    // enable if needed
+    // npmOptions: ['--config.strict-peer-dependencies=false'],
```
Techn1x, line 10: Things are stable enough in the try scenarios we can make the peers strict again

PR #546, opened 2025-05-19
chore: improve pnpm config, fix ember-release try scenario

Techn1x:
Noticed that we don't even need ember-source-channel-url, it seems other repos get away without it just fine, eg https://github.com/ember-modifier/ember-modifier/pull/950


Techn1x:
`ember-release` scenario does seem to succeed now, though I am going to leave it as "allowed to fail" in the CI in case later ember versions cause issues

PR #549, opened 2025-06-20
feat: init RequestManagerMSW

Techn1x:


addon/src/mocks/request-manager-msw.js
```diff
@@ -0,0 +1,65 @@
...
+
+    if (!wrapper) {
+      wrapper = new RequestWrapper(); // this generates & returns the handler function
+      const mswHandler = http[type.toLowerCase()].call(
+        this.msw,
+        url,
+        async ({ request, params }) => {
+          await delay(this._settings.delay);
+          return await wrapper({ request: request.clone(), params });
+        },
```
Techn1x, line 49: native stuff makes this easy

addon/src/mocks/request-manager-pretender.js
```diff
@@ -52,7 +52,37 @@ export default class RequestManagerPretender extends RequestManager {
...
+            params: fakeRequest.params,
+          });
+          if (!response) return;
+
+          // Turn Response into the response array that pretender expects
+          const responseHeaders = {};
+          response.headers.forEach(
+            (value, key) => (responseHeaders[key] = value),
+          );
+          return [response.status, responseHeaders, await response.text()];
```
Techn1x, line 84: handler for pretender - requires a lot more handling because of its shenanigans, now that the rest of the addon is expecting native Request/Response etc.

addon/src/index.js
```diff
@@ -65,7 +67,14 @@ export function setupFactoryGuy(hooks, requestManager) {
...
 
-    const rm = requestManager ? requestManager : new RequestManagerPretender();
+    let DefaultRequestManager;
+    if (interceptor === 'pretender') {
+      DefaultRequestManager = RequestManagerPretender;
+    } else if (interceptor === 'msw') {
+      DefaultRequestManager = RequestManagerMSW;
+    }
+
+    const rm = requestManager ? requestManager : new DefaultRequestManager();
```
Techn1x, line 77: request manager can be provided to the setupFactoryGuy() function, otherwise it will default to what it set in the config

addon/src/mocks/mock-request.js
```diff
@@ -1,18 +1,12 @@
 import { assert } from '@ember/debug';
-import { typeOf } from '@ember/utils';
-import {
-  isEmptyObject,
-  isEquivalent,
-  isPartOf,
-  param,
-  toParams,
-} from '../utils/helper-functions';
```
Techn1x, line 9: aimed to reduce the need for bespoke code / utility functions, makes it a lot harder for new maintainers to know exactly what each function is supposed to do, without spending lots of time learning them

addon/src/mocks/mock-request.js
```diff
@@ -139,55 +151,94 @@ export default class {
...
+      // request must have every specified item in someQueryParams, but can have more.
+      const someQueryParams = new URLSearchParams(param(this.someQueryParams)); // convert json -> URLSearchParams
+      return someQueryParams
+        .keys()
+        .every((key) =>
+          isEqual(
+            someQueryParams.getAll(key),
+            requestUrl.searchParams.getAll(key),
+          ),
+        );
```
Techn1x, line 167: largely the same, just trying to use more native javascript to do the heavy lifting. In this case `URL` and `URLSearchParams` provide us a bunch for free

addon/src/mocks/mock-request.js
```diff
@@ -139,55 +151,94 @@ export default class {
...
+    }
+    try {
+      return await request.clone().formData();
+    } catch (e) {
+      // continue
+    }
+    try {
+      return await request.clone().text();
+    } catch (e) {
+      // continue
```
Techn1x, line 213: we can't really rely on request headers being set in order to determine this, so we just try each one.

addon/src/mocks/request-manager-msw.js
```diff
@@ -0,0 +1,70 @@
...
+      return;
+    }
+
+    if (macroCondition(dependencySatisfies('msw', '*'))) {
+      const { setupWorker } = importSync('msw/browser');
+      msw = new setupWorker();
+      this.msw = msw;
+    } else {
+      throw new Error('RequestManagerMSW needs msw dependency');
+    }
```
Techn1x, line 32: only import msw if actually needed

addon/src/mocks/request-wrapper.js
```diff
@@ -49,12 +49,13 @@ export default class RequestWrapper {
...
-   *
+   * Sort the handlers by those with queryParams first, then matchArgs second, then anything else
    */
   getHandlers() {
-    return this.handlers.sort(
-      (a, b) => b.hasQueryParams() - a.hasQueryParams(),
+    return [...this.handlers].sort(
+      (a, b) =>
+        b.hasQueryParams() - a.hasQueryParams() ||
+        Boolean(b.matchArgs) - Boolean(a.matchArgs),
```
Techn1x, line 58: allows for more gauranteed ordering of which handler matches a request, regardless of which order they are added
```ts
mockCreate('thing') // third
mockCreate('thing').withParams(...) // first
mockCreate('thing').matches(...) // second
```

addon/src/mocks/request-wrapper.js
```diff
@@ -68,37 +69,16 @@ export default class RequestWrapper {
...
-    let handler = this.getHandlers(request).find((handler) =>
-      handler.matches(request),
-    );
-    if (handler) {
-      let { status, headers, responseText } = handler.getResponse();
-      return [status, headers, responseText];
+  async handleRequest({ request, params }) {
+    for (const handler of this.getHandlers()) {
+      if (await handler.matches({ request, params })) {
+        return handler.getResponse();
```
Techn1x, line 80: match on a handler, return its response! it's that easy 😛

test-app/ember-cli-build.js
```diff
@@ -6,6 +6,7 @@ module.exports = async function (defaults) {
       setConfig: {
         'ember-data-factory-guy': {
           useStringIdsOnly: true,
+          interceptor: process.env.INTERCEPTOR || 'pretender',
```
Techn1x, line 9: this shows how you can configure which interceptor to use application wide

We have ember-try modify this so that we can run the same test suite against both Pretender and MSW

test-app/tests/unit/shared-adapter-behaviour.js
```diff
@@ -1149,19 +1149,18 @@ SharedBehavior.mockCreateTests = function () {
...
-      .save()
-      .catch(() => {
-        assert.ok(true);
-        // our mock was NOT called
-        assert.strictEqual(mock.timesCalled, 0);
-      });
+      .save();
+
+    assert.strictEqual(baseMock.timesCalled, 1);
+    assert.strictEqual(otherMock.timesCalled, 0);
```
Techn1x, line 1163: use of `.catch()` no longer catches the errors on these when the handler misses. I couldn't figure out a way to catch the errors thrown there

Ultimately it's because the request handler function is now async, and Pretender treats async handlers differently - it doesn't throw, `erroredRequest` doesn't get called
https://github.com/pretenderjs/pretender/blob/65fbf0a608f2e4b4fd79657a77a41f52077074a9/src/pretender.ts#L238-L242

Instead, I've just made sure to assert `timesCalled` is as expected

test-app/tests/unit/shared-adapter-behaviour.js
```diff
@@ -1419,21 +1425,10 @@ SharedBehavior.mockUpdateTests = function () {
 
       let fd = new FormData();
       fd.append('description', snapshot.attr('description'));
-      adapter.ajax(url, httpMethod, {
-        data: fd,
-        processData: false,
-        contentType: false,
-      });
+      const f = await fetch(url, { method: httpMethod, body: fd });
+      return f.text();
```
Techn1x, line 1429: use native fetch instead here

PR #562, opened 2026-05-05
update release-plan workflows

mansona:
Hey 👋  I noticed in https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/561 that you updated one of the workflows and I saw that you weren't on the latest iteration of the release-plan stuff

I tend to update the release-plan workflows using `pnpm create release-plan-setup@latest` to make sure that I am on the latest (we put a lot of work into those templates to make sure people get the best experience)

This PR is just the output of that command 👍

### DIFF
--- a/.github/workflows/plan-release.yml
+++ b/.github/workflows/plan-release.yml
@@ -1,5 +1,6 @@
-name: Release Plan Review
+name: Plan Release
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


mfeckie:
Great, was getting ready to start reading how release plane worked

mansona:
It should all be setup right here 👍 all you need to do is label the unlabeled PR in https://github.com/adopted-ember-addons/ember-data-factory-guy/pull/555 and it will show you what the changelog will look like

PR #563, opened 2026-05-05
Fix bugs and design issues found in audit

mfeckie:
I ran an automated audit of the codebase and found a bunch of inconsistencies that seem worth fixing.

Mostly Claude, but each commit was reviewed and tested by me.

Basically I'm looking at getting this addon all up to date so it's in a good place to iterate on ensuring it works with https://warp-drive.io/

## Summary

- **`super.setttings` typo** in `RequestManagerPretender` — would throw a TypeError at runtime when `.settings()` was called
- **`mockLinks` assert silently ignored `relationshipKey`** — `assert(msg, model, relationshipKey)` passes 3 args; Ember's `assert` only takes 2, so the second condition was swallowed
- **`lookupDefinitionForFixtureName` inverted assert condition** — `!definition && assertItExists` was always false when any definitions existed, causing the assert to fire in the wrong cases; fixed to `!assertItExists`
- **`mockReload` type guard always true** — `typeof parseInt(x) === 'number'` is true even for `NaN`; replaced with `args[1] != null`
- **`cacheOnlyMode` patched `store.adapterFor` permanently** — the override was never cleaned up in `reset()`, leaking into subsequent tests; `reset()` now restores the original
- **`Sequence` was a constructor function, not a class** — `fn.call(this, index++)` bound `this` to the Sequence instance inside user-supplied callbacks; converted to a proper `class` with `fn` called without a receiver
- **Anonymous `export default class`** in `MockRequest`, `MockStoreRequest`, `Scenario` — named so they appear correctly in stack traces and `constructor.name` checks
- **`RequestWrapper` constructor returned a function** — meant `instanceof RequestWrapper` always returned `false`; removed the return so the constructor behaves normally

## Test plan

- [x] All existing tests still pass (1001/1001)
- [x] New unit tests added for each fix: `RequestManagerPretender`, `mockLinks` guard message, `lookupDefinitionForFixtureName`, `mockReload` guard, `cacheOnlyMode` reset, `Sequence` class behaviour, `MockRequest`/`MockStoreRequest` class names, `RequestWrapper` instanceof

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/addon/src/-private.js
+++ b/addon/src/-private.js
@@ -2,6 +2,8 @@
 
 import FixtureBuilderFactory from './builder/fixture-builder-factory';
 import MockStoreRequest from './mocks/mock-store-request';
+import RequestWrapper from './mocks/request-wrapper';
+import Sequence from './sequence';
 
 import {
   isEmptyObject,
@@ -13,6 +15,8 @@ import {
 export {
   FixtureBuilderFactory,
   MockStoreRequest,
+  RequestWrapper,
+  Sequence,
   isEmptyObject,
   param,
   isEquivalent,
--- a/addon/src/factory-guy.js
+++ b/addon/src/factory-guy.js
@@ -592,7 +592,7 @@ class FactoryGuy {
 
     assert(
       `[ember-data-factory-guy] Can't find that factory named [ ${name} ]`,
-      !definition && assertItExists,
+      !assertItExists,
     );
   }
 
--- a/addon/src/factory-guy.js
+++ b/addon/src/factory-guy.js
@@ -450,6 +450,10 @@ class FactoryGuy {
   }
 
   reset() {
+    if (this._cacheOnlyOriginalAdapterFor && this.store) {
+      this.store.adapterFor = this._cacheOnlyOriginalAdapterFor;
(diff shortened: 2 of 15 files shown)


mansona:
Hey @mfeckie sorry I don't really have time to review this kind of PR right now 🙈 so please don't wait for me (or try to find someone else to review it for you)

Essentially since this PR is doing a whole bunch of changes all at once, I would need to onboard so much context to know if the change is good or bad. I tend to [recommend PRs with a "single logical change"](https://mainmatter.com/blog/2021/05/26/keeping-a-clean-git-history/) rather than big ones like this, not necessarily for a glean git history but just because it's much easier/faster to review

mfeckie:
Fair enough. I'm confident that the surface area of change is actually small and fixes only bugs without changing external behaviour. Most of the stuff is added tests.

I'm trying to work towards making the library compatible with WarpDrive and so wanted to update dependencies and do a tidy up so things are in a 'pristine' condition before trying to adapt for WarpDrive.

On that basis, the tests are all passing and I think this is genuinely low risk, I'll merge the PR and start a more robust process for the steps to get compatible with warp drive.

